PR #5, opened 2017-02-03
Fix t.object use-case

dmorosinotto:
Hi @gcanti, great idea and amazing work! 
Reading the docs, I think that I have spotted a simple typo in the t.object definition use-case, see the fix in this PR.
And also I suggest to use the "annotation" specification in the header of the Typescript syntax, and maybe it can be completed adding a ":" in front of all the rows in the table (in that column) I don't know what you think about it...


gcanti:
Good catch, thanks @dmorosinotto

PR #52, opened 2017-06-08
Tagged Union of Types, closes #50

### LINKED ISSUE #50: Tagged Union of Types
I really love this library and was hoping to be able to write something similar to Haskell's generic functions. But to leverage totality checking I need to be able to form some kind of tagged union of the Types in this package. But I'm having some trouble.

I basically want something like this, where I can pass a Type object and a map of folds to handle each type of Type to build database schemas and stuff from the Type definition.

Any help would be greatly appreciated.

```
const assertNever = (x: never): never => {throw new Error(`Unexpected object ${x}`)}

type Types<A = any, B extends Props = {}, C extends Any[] = Any[], D = any> =
    | Type<A>
    | InterfaceType<B>
    | UnionType<C, D>

function genericFold<A = any, B extends Props = {}, C extends Any[] = Any[], D = any>(x : Types<A, B, C, D>, f: any) {
    if (x instanceof UnionType) {
        // Do something
        if (x.t instanceof Type) genericFold(x.t);
    } else if (x instanceof InterfaceType) {

    } else if (x instanceof Type) {

    } else {
        return assertNever(x);
    }
}
```

gcanti:
This PR moves `io-ts` to structural typing removing all `class`es. Here's the list of breaking changes so far

- `Type` is now an interface
- types no more own a `is` method, use `t.is` instead
- unions no more own a `fold` method
- `Reporter`, `PathReporter`, `ThrowReporter` are now top level modules

@justinwoo Could you please try it out? Does it break badly some of your code?

`lib` is committed in so you can install it by running

```
npm i gcanti/io-ts#50
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 *.log
 node_modules
 lib
-lib-jsnext
 dev
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,14 @@
 **Note**: Gaps between patch versions are faulty/broken releases.
 **Note**: A feature tagged as Experimental is in a high state of flux, you're at risk of it changing without notice.
 
+# 0.5.0
+
+- **Breaking Change**
+  - `Type` is now an interface
+  - types no more own a `is` method, use `t.is` instead
+  - unions no more own a `fold` method
+  - `Reporter`, `PathReporter`, `ThrowReporter` are now top level modules
+
 # 0.4.0
 
 - **Breaking Change**
--- a/README.md
+++ b/README.md
@@ -3,9 +3,10 @@
 A value of type `Type<T>` (called "runtime type") is the runtime representation of the static type `T`:
 
 ```js
-class Type<T> {
-  constructor(public readonly name: string, public readonly validate: Validate<T>) {}
-  is(x: any): x is T
+export interface Type<A> {
+  readonly _A: A
+  readonly name: string
+  readonly validate: Validate<A>
 }
 ```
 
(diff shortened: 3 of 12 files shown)


justinwoo:
I didn't see any problems, gives me correct signatures and all 👍

vegansk:
It brokes my tool that creates typed paths for react-redux-forms controls :-( Will investigate what's wrong with it

vegansk:
I guess I just need to check the `_tag` field instead of using the `instanceof` operation
https://gist.github.com/vegansk/94668bf2c1746bd3ebcc41ce7e6a77ee

gcanti:
@vegansk There's more than that: the `Type` interface doesn't own a `_tag` field, therefore you can't pattern match, you'd need a union (see the linked issue for an example)

@justinwoo Nice, thanks!

sledorze:
@gcanti just wondering if that could make the typescript compiler crazier (you know the deep comparison of Union types resulting in memory exhaustion)..

gcanti:
@sledorze I think (hope?) that the issue with memory exhaustion doesn't apply here: as far as I know that issue was due to a combination of function overloadings, accepting unions of classes, with lots of methods. In this refactoring classes (and methods) are gone and we end up with just raw tagged unions of POJOs

sledorze:
@gcanti oh good to hear, I hope too.

gcanti:
@sledorze or you are talking about the other issue, i.e. when tsc crashes? That seems related to multiple versions in `node_modules` (https://github.com/Microsoft/TypeScript/issues/16029#issuecomment-304991151)

gcanti:
@vegansk ping: any news?

vegansk:
@gcanti , it's ok for me :-)

gcanti:
> you can install it by running `npm i gcanti/io-ts#50`

I'm going to remove the `lib` folder and merge this PR, if anyone is depending on this branch, please make a clone for the time being.

PR #85, opened 2017-11-05
add `strict` combinator, closes #84

### LINKED ISSUE #84: Ceiling Type / Strict Interface
I think a type that specifies that only the given interface properties are allowed would be useful. Working Implementation:

```ts
export class CeilingType<RT extends t.InterfaceType<any>> implements t.Type<t.TypeOf<RT>> {
	readonly _tag: "CeilingType" = "CeilingType";
	readonly _A: t.TypeOf<RT>;
	readonly validate: t.Validate<t.TypeOf<RT>>;
	constructor(readonly type: RT, readonly name: string = `ceil(${type.name})`) {
		const shouldBeProps = Object.getOwnPropertyNames(type.props);
		this.validate = (v, c) =>
			type.validate(v, c).chain(obj => {
				const props = Object.getOwnPropertyNames(obj);
				if (props.length === shouldBeProps.length) return t.success(obj);
				for (const prop of props) {
					if (!shouldBeProps.includes(prop))
						return t.failure(obj[prop], [...c, { key: prop, type: t.never }]);
				}
				throw `Impossible? got [${props.join(", ")}], expected  [${shouldBeProps.join(", ")}]`;
			});
	}
}
export function strictInterface<P extends t.Props>(props: P, name?: string): CeilingType<t.InterfaceType<P>> {
	return new CeilingType(t.interface(props, name));
}
```

Not sure what it should be called, so far i came up with "ceiling type" or "strict interface".

This implementation only works for own properties and ignores symbols, and there might be some special handling for unenumerable properties needed.

gcanti:


README.md
```diff
@@ -162,6 +162,7 @@ import * as t from 'io-ts'
 | refinement | ✘ | `t.refinement(A, predicate)` |
 | map | ✘ | `t.map(f, type)` |
 | prism | ✘ | `t.prism(type, getOption)` |
+| strict | ✘ | `t.strict(type)` |
```
phiresky: There is no mention that this only works on interfaces, and what it means exactly.

src/index.ts
```diff
@@ -583,4 +585,35 @@ export class ReadonlyArrayType<RT extends Any> implements Type<ReadonlyArray<Typ
...
 
+export class StrictType<RT extends InterfaceType<any>> implements Type<TypeOf<RT>> {
+  readonly _tag: 'StrictType' = 'StrictType'
+  readonly _A: TypeOf<RT>
+  readonly validate: Validate<TypeOf<RT>>
+  constructor(readonly type: RT, readonly name: string = `StrictType<${type.name}>`) {
+    const len = Object.keys(type.props).length
+    this.validate = (v, c) =>
+      type.validate(v, c).chain(o => {
+        const keys = Object.keys(v)
```
phiresky: I'm not sure if Object.keys is the correct function here? I'm not certain how symbols and non-enumerable properties should be handled. This is enough for me as all my data is deserialized from JSON before, so it can't contain any symbols or non-enumerable properties, but it might not be enough for the generic case.

phiresky: Also, shouldn't it be `Object.keys(o)` instead of `Object.keys(v)`?

src/index.ts
```diff
@@ -583,4 +585,35 @@ export class ReadonlyArrayType<RT extends Any> implements Type<ReadonlyArray<Typ
...
+          const errors: Errors = []
+          for (let i = 0; i < keys.length; i++) {
+            const key = keys[i]
+            if (!type.props.hasOwnProperty(key)) {
+              errors.push(getValidationError(o[key], c.concat(getContextEntry(key, never))))
+            }
+          }
+          if (errors.length) {
+            return failures(errors)
+          }
```
phiresky: This means the type check succeeds on some unknown case where the keys length is not the expected length, right?

I don't think I'm comfortable with letting it succeed in that case because it means something is wrong, could you make it fail with just the outer error (whole interface does not match)?

src/index.ts
```diff
@@ -583,4 +585,35 @@ export class ReadonlyArrayType<RT extends Any> implements Type<ReadonlyArray<Typ
...
+        }
+        return success(o)
+      })
+  }
+}
+
+/** Specifies that only the given interface properties are allowed */
+export function strict<RT extends InterfaceType<any>>(type: RT, name?: string): StrictType<RT> {
+  return new StrictType(type, name)
+}
```
phiresky: Because strict only works with interfaces anyways, i think it might be clearer (and more compact) for this function to directly take Props as an argument instead of an interface.

```ts
export function strict<P extends Props>(props: P, name?: string): StrictType<InterfaceType<P>> {
  return new StrictType(type(props), name)
}
```

README.md
gcanti: I thought that the signature and the comment "specifies that only the given interface properties are allowed" would be enough, feel free to suggest any improvement though

src/index.ts
gcanti: > This is enough for me

For me too

> shouldn't it be Object.keys(o)

Right, thanks

src/index.ts
gcanti: Good idea

README.md
phiresky: Right, just saw you added a JSDoc comment to the actual function. That's probably enough. You could write ``t.strict(interface)` though.

src/index.ts
gcanti: Agreed, also is more consistent with the `partial` combinator

gcanti:
@phiresky Ok, these are the changes I made so far

- use `Object.getOwnPropertyNames` instead of `Object.keys`
- if `keys.length !== len` make sure the validation will fail
- use `o` instead of `v` while retrieving the properties
- accept a map of props instead of a type
- added an example to the readme

let me know if it's ok

phiresky:
Looks fine to me!

gcanti:
Nice, thanks for your help.

gcanti:
Released

PR #89, opened 2017-11-17
add Flow definition file

gcanti:


aaronjensen:
Hey @gcanti, thanks for this. We're giving this a shot on our project. I had one question, I apologize if it's obvious, but how would I specify that something is a `Date`?

aaronjensen:
Also, could we have a `t.mixed` please, or is that possible now?

gcanti:
@aaronjensen it depends on the source format, a couple of examples: (1) a date from millis  and (2) a date from an iso string 

```ts
const DateFromMillis = new t.Type<any, Date>(
  'DateFromMillis',
  (v): v is Date => v instanceof Date,
  (v, c) =>
    t.number.validate(v, c).chain(n => {
      const d = new Date(n)
      return isNaN(d.getTime()) ? t.failure(n, c) : t.success(d)
    }),
  v => v.getTime()
)

const DateFromISOString = new t.Type<any, Date>(
  'DateFromISOString',
  (v): v is Date => v instanceof Date,
  (v, c) =>
    t.string.validate(v, c).chain(s => {
      const d = new Date(s)
      return isNaN(d.getTime()) ? t.failure(s, c) : t.success(d)
    }),
  v => v.toISOString()
)
```

aaronjensen:
I'm running into some unexpected issues:

```js
export type APIEndpoint<Response> = {
  url: string,
  type: t.Type<any, Response>,
}

const apiEndpoint = (url, type) => ({ url, type })

export const tId = t.union([t.string, t.number])
export type Id = t.TypeOf<typeof tId>

export const tTeam = t.type({
  id: tId,
  name: t.string,
})
export type Team = t.TypeOf<typeof tTeam>

export const teamPath = (id: Id): APIEndpoint<Team> =>
  apiEndpoint(`/teams/${id}`, tTeam)
```


```
$ /Users/aaronjensen/Source/layer/project/node_modules/.bin/flow
Error: node_modules/io-ts/lib/index.js.flow:220
220: declare export class ArrayType<RT: Any> extends Type<any, Array<TypeOf<RT>>> {
                                                               ^^^^^^^^^^^^^^^^^ array type. This type is incompatible with
 31: export const teamPath = (id: Id): APIEndpoint<Team> =>
                                                   ^^^^ object type. See: packages/frontend/src/api/paths.js:31

Error: node_modules/io-ts/lib/index.js.flow:242
242: declare export class InterfaceType<P: Props> extends Type<any, InterfaceOf<P>> {
                                                                    ^^^^^^^^^^^^^^ object type. This type is incompatible with
 31: export const teamPath = (id: Id): APIEndpoint<Team> =>
                                                   ^^^^ object type. See: packages/frontend/src/api/paths.js:31
  Property `name` is incompatible:
     31: export const teamPath = (id: Id): APIEndpoint<Team> =>
                                                       ^^^^ property `name`. Property not found in. See: packages/frontend/src/api/paths.js:31
    242: declare export class InterfaceType<P: Props> extends Type<any, InterfaceOf<P>> {
                                                                        ^^^^^^^^^^^^^^ object type
```

It may need more from my project to repro, but perhaps you see a problem w/ what I've done already

aaronjensen:
It looks like this was an inference issue. This seems to work:

```
const apiEndpoint = <R>(url, type: t.Type<any, R>): APIEndpoint<R> => ({
  url,
  type,
})

export const teamPath = (id: Id): APIEndpoint<Team> =>
  apiEndpoint(`/teams/${id}`, tTeam)
```

aaronjensen:
Ok, I've successfully added this to validate our api fetching and it's awesome. Thank you for this! 

I know you're indicated that this is experimental, but do you have plans to support this and the TypeScript version going forward?

gcanti:
@aaronjensen Great, thanks for testing out.

> do you have plans to support this and the TypeScript version going forward

Yes, my intention is adding Flow support to

- `fp-ts` (which `io-ts` depends upon, for the `Either` type)
- `io-ts`
- `monocle-ts`

I'm trying to build a good ecosystem around fp-ts and would be nice if Flow users too can profit from those libraries

aaronjensen:
@gcanti that's great news.

One piece of feedback, I actually thought that read only array support wasn't in the build you released because I tried to spell it `t.readOnlyArray` (flow's is `$ReadOnlyArray`, capital `O`). I looked again and realized it's a lower case o here. I don't know if that's a typescript-ism, but I thought I'd point it out as it was a source of confusion. Same applies to `$ReadOnly` vs `readonly`.

Also, this is what I've got for `mixed`, any problems w/ it?

```js
export const tMixed: t.Type<any, mixed> = new t.Type(
  'mixed',
  () => true,
  m => t.success(m),
  m => m,
)
```

gcanti:
> I don't know if that's a typescript-ism

Ah yes, in TypeScript there is `Readonly` and `ReadonlyArray`. I added a column "Flow" in the readme.

> Also, this is what I've got for mixed

What's the use case for that?

aaronjensen:
> What's the use case for that?

Good question. It was part of a 3rd party lib I was typing. It was something I needed to pass opaquely. I ended up not actually needing it, so I pulled it out. I also realize now that typing `mixed` with `io-ts` is against the point of `io-ts` 😄

gcanti:
@aaronjensen I released a new `io-ts@next` version which imports the shadow files contained in `fp-ts@next`

aaronjensen:
Ok thanks I will give it a shot. Are you planning to keep your releases in sync with flow versions since you are using shadow files rather than flow-typed?

aaronjensen:
Fwiw I am fine with that since we always tend to keep our flow up to date.

gcanti:
I guess so. Do you think will be a burden? io-ts's APIs are quite stable

aaronjensen:
> I guess so. Do you think will be a burden? io-ts's APIs are quite stable

Ah, so if io-ts's apis don't change then you can use the versions only for ts/flow upgrades. It's unfortunate that there's no way to specify flow versions that it supports in package.json. 

Is there a place to see upcoming TS 2.7 changes?

gcanti:
The roadmap https://github.com/Microsoft/TypeScript/wiki/Roadmap (and sometimes the PRs)

PR #92, opened 2017-11-28
docs: cache NODE_ENV in unsafe validate example

dvlsg:
Just a small documentation update, if you're open to it. Suggesting that `NODE_ENV` be "cached" in the example `unsafeValidate` method.

Reasoning / discussions here:
* https://github.com/nodejs/node/issues/3104
* https://github.com/nodejs/node/pull/1648

Unless you'd expect `NODE_ENV` to be editable while the application is running (debugging on a live server, perhaps?)


gcanti:
Thanks @dvlsg

PR #99, opened 2017-12-11
Various optimizations

gcanti:
Using `./perf/index.js`

---

`concat(element)` (current implementation)

invalid payload x 200,425 ops/sec ±0.74% (89 runs sampled)
valid payload (no errors) x 181,576 ops/sec ±1.00% (87 runs sampled)

`concat([element])`

invalid payload x 258,902 ops/sec ±0.61% (89 runs sampled)
valid payload (no errors) x 227,362 ops/sec ±1.54% (86 runs sampled)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "6"
+  - "8"
 before_script:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
       "prettier --no-semi --single-quote --print-width 120 --parser flow --write \"{src,test,examples,exercises}/**/*.js.flow\"",
     "test": "npm run prettier && npm run lint && npm run typings-checker && npm run mocha",
     "clean": "rm -rf lib/*",
-    "build": "npm run clean && tsc && npm run flow-copy-definition-files"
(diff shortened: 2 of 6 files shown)


gcanti:
@sledorze this looks a real gain though

gcanti:
It's pretty surprising by the way, I'm doing something wrong?

This is another benchmark

```ts
var Benchmark = require('benchmark')

const suite = new Benchmark.Suite()

const x = [1, 2, 3]

suite
  .add('concat(a)', function() {
    x.concat(4)
  })
  .add('concat([a])', function() {
    x.concat([4])
  })
  .on('cycle', function(event) {
    console.log(String(event.target))
  })
  .on('complete', function() {
    console.log('Fastest is ' + this.filter('fastest').map('name'))
  })
  .run({ async: true })
```

Results

```
x.concat(4) // 2,935,475 ops/sec ±0.87% (82 runs sampled)
x.concat([4]) // 5,421,103 ops/sec ±0.93% (86 runs sampled)
```

sledorze:
@gcanti that's crazy, maybe due to recent optimisations changes in V8.
I'm wondering if, as we decide to choose an implementation over another based on a comparison, if we should not enforce in a test that, so that if it ever change, we can adapt..

`x.concat([4])` must be faster than `x.concat(4)`

gcanti:
> if we should not enforce in a test that, so that if it ever change, we can adapt..

That would be nice

PR #101, opened 2017-12-12
add Decoder / Encoder interfaces

gcanti:


src/index.ts
```diff
@@ -9,7 +9,7 @@ declare global {
 
 export interface ContextEntry {
   readonly key: string
-  readonly type: Any | NeverType
+  readonly type: Decoder<any, any>
```
sledorze: @gcanti do we loose anything by dropping 'never' ?

gcanti: @sledorze `| NeverType` was added for variance issues, IMO this is an actual improvement due to the Decoder/Encoder splitting

sledorze: @gcanti  ok good to know 👍

gcanti: For reference, these are the lines that currently cause variance issues if I remove `| NeverType`

https://github.com/gcanti/io-ts/blob/37c74a5e2038de063a950f9ba8d18b1f132ef450/src/index.ts#L756
https://github.com/gcanti/io-ts/blob/37c74a5e2038de063a950f9ba8d18b1f132ef450/src/index.ts#L875

gcanti:
@sledorze rebased on top of master, should I release a next version after merging this?

sledorze:
@gcanti yes please, I may integrate it over le week-end.

gcanti:
@sledorze published as `io-ts@0.9.2-dev.20171215`. Let me know how it goes

sledorze:
@gcanti it works perfectly! The code involved looks nicer too!
(Note: I had to manually change io-ts-types ref to io-ts to prevent type duplication)

PR #103, opened 2017-12-13
add appendContext

gcanti:
@sledorze with this change we go from

(branch `perf`)

invalid payload x 262,497 ops/sec ±1.00% (85 runs sampled)
valid payload (no errors) x 308,716 ops/sec ±1.98% (86 runs sampled)

to

(branch `perf-appendContext`)

invalid payload x 511,440 ops/sec ±1.17% (87 runs sampled)
valid payload (no errors) x 686,213 ops/sec ±0.79% (86 runs sampled)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,6 @@
 import { Either, Left, Right, isRight } from 'fp-ts/lib/Either'
 
 import { Predicate } from 'fp-ts/lib/function'
-import { snoc } from 'fp-ts/lib/Array'
 
 declare global {
   interface Array<T> {
@@ -64,6 +63,16 @@ export const getContextEntry = (key: string, type: Any | NeverType): ContextEntr
 export const getValidationError = (value: any, context: Context): ValidationError => ({ value, context })
 
 export const getDefaultContext = (type: Any): Context => [{ key: '', type }]
+
+export const appendContext = (c: Context, key: string, type: Any | NeverType): Context => {
+  const len = c.length
+  const r = Array(len + 1)
+  for (let i = 0; i < len; i++) {
+    r[i] = c[i]
+  }
+  r[len] = { key, type }
+  return r
+}
 
 export const failures = <T>(errors: Errors): Validation<T> => new Left(errors)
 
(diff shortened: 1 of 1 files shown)


sledorze:
@gcanti simply great 👍

gcanti:
@sledorze I am... puzzled. Look at this benchmark (I'm trying to optimize error concatenation now)

```ts
var Benchmark = require('benchmark')

const suite = new Benchmark.Suite()

const pushAll1 = (xs, ys) => Array.prototype.push.apply(xs, ys)

const concat = (xs, ys) => {
  const lenx = xs.length
  const leny = ys.length
  const r = Array(lenx + leny)
  for (let i = 0; i < lenx; i++) {
    r[i] = xs[i]
  }
  for (let i = 0; i < leny; i++) {
    r[i + lenx] = ys[i]
  }
  return r
}

suite
  .add('pushAll1', function() {
    pushAll1([1, 2, 3], [4, 5])
  })
  .add('concat', function() {
    concat([1, 2, 3], [4, 5])
  })
  .on('cycle', function(event) {
    console.log(String(event.target))
  })
  .on('complete', function() {
    console.log('Fastest is ' + this.filter('fastest').map('name'))
  })
  .run({ async: true })
```

Results

```
pushAll1 x 14,376,146 ops/sec ±0.48% (87 runs sampled)
concat x 20,219,513 ops/sec ±0.65% (85 runs sampled)
```

What does this mean? Is immutable concatenation *faster* than mutable push?

sledorze:
@gcanti I tend to think maybe 'native' calls go through the Javascript -> C++ Layer.
so it may indeed be faster not to cross it.
HOWEVER, we're creating more garbage (I dunno if the GC has some policy for those very short lived objects)

gcanti:
> maybe a basic cons (lisp like) structure could even be faster?

We can try as an exercise (or for the next major release), it would be a breaking change

> HOWEVER, we're creating more garbage

Yeah, I guess we should leave the error handling as it is (mutable `pushAll`)

sledorze:
@gcanti if breaking change is allowed then maybe even not doing all that error structure upfront but only in case of an error could be faster (on the return path).

PR #109, opened 2018-01-18
Replace any with mixed

gcanti:
@sledorze `io-ts`'s APIs are pretty stable, this is the last change (*) I would like to make before v1.0.0, what do you think?

p.s.
Actually I'm not sure if is breaking

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ Also a runtime type can
 * be used as a custom type guard (through `is`)
 
 ```ts
+export type mixed = object | number | string | boolean | symbol | undefined | null
+
 class Type<S, A> {
   readonly _A: A
   readonly _S: S
@@ -16,7 +18,7 @@ class Type<S, A> {
     /** a unique name for this runtime type */
     readonly name: string,
     /** a custom type guard */
-    readonly is: (v: any) => v is A,
+    readonly is: (v: mixed) => v is A,
     /** succeeds if a value of type S can be decoded to a value of type A */
     readonly validate: (input: S, context: Context) => Either<Errors, A>,
     /** converts a value of type A to a value of type S */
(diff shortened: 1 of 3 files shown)


sledorze:
@gcanti I'm totally in favor of stricter typings.
I think that if it breaks something then the error would/should be addressed in the related code and not by narrowing back this type to any.

Why using 'Std' name and what does it means?
(potentially so used that it would potentially lead to many naming collision).

Issue related to our experience with io-ts for 1.0; this one could be very annoying:
https://github.com/gcanti/io-ts/issues/110

gcanti:
> I think that if it breaks something then the error would/should be addressed in the related code 

Yeah, theoretically if it breaks something should mean that there's a bug or some unsafe code in the consumer

> Why using 'Std' name and what does it means?

Oh, it was just a private type alias in the Flow definition file

```ts
type Std<A> = Type<mixed, A>
```

It's pretty useless though, so I'll remove it

gcanti:
Well this change seems to work, I found a fragment of unsafe code in fp-ts-routing..

sledorze:
@gcanti i ll apply that to our codebase soon!

gunzip:
hi @gcanti , it looks this was a breaking change for us. 

ie. I get 

```
Il tipo 'Validate<{}, InterfaceOf<{ markdown: Type<any, string & IWithinRangeStringTag<80, 10000>>; }> & P...' non è assegnabile al tipo 'Validate<mixed, any>'.
          I tipi dei parametri 's' e 's' sono incompatibili.
            Il tipo 'mixed' non è assegnabile al tipo '{}'.
              Il tipo 'undefined' non è assegnabile al tipo '{}'.
```

with this code:

https://github.com/teamdigitale/digital-citizenship-functions/blob/master/lib/models/message.ts#L58

which is an intersection of two interfaces. 

may you suggest the best way to fix it in our code ?

PR #115, opened 2018-01-26
make recursion safer

gcanti:
Currently the following doesn't raise any error

```ts
// string???
const T = t.recursion<string>('T', Self =>
  t.interface({
    type: t.literal('a'),
    items: t.array(Self)
  })
)
```


gcanti:
Closing in favour of https://github.com/gcanti/io-ts/pull/117

PR #118, opened 2018-01-30
 add validate overload with implicit context

mmkal:
this adds an overload to `validate` which takes a single argument - a POJO that is expected to be the target type. It doesn't take a context array, and it does compile-time validation of the input type.

This is useful if you want to build up an interface instance out of variables from external sources, and you want:
1. the IDE to be able to give you type hints of what properties *should* be on the object (see VSCode example below)
2. to be able to use the more natural style of `Person.validate({ name: "Bob" })` syntax over `t.validate({ name: "Bob" }, Person)`
3. to not have to pass in a context array explicitly like `Person.validate({ name: "Bob" }), [{ key: "", type: Person }])` or more likely (but worse), no context at all like `Person.validate({ name: "Bob" }, [])`
4. to not have to `import * as t from "io-ts"` when you just want to use types defined elsewhere, but don't want to import io-ts directly. You can just `import { Person } from "./foo/bar"`

Usage example:
```javascript
import * as t from "io-ts"

const Person = t.interface({
  name: t.string,
  age: t.number,
})

const alice = Person.validate({ name: "alice", age: "thirty" }) // compile time error, age should be of type "number"
const bob = Person.validate({ name: "bob", age: 30 }) // equivalent of calling t.validate({ name: "bob", age: 30 }, Person)
```

VSCode type hints:
image

Curious to see what you think @gcanti - for the application I'm trying to use this in, it'd make io-ts a no-brainer as it'll give a very clean way to get both runtime and compile time type-checking.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -49,18 +49,23 @@ export class Type<S, A> implements Decoder<S, A>, Encoder<S, A> {
   readonly '_A': A
   // prettier-ignore
   readonly '_S': S
-  readonly maybe: (value: A) => Validation<A>
+  readonly validate: {
+    /** succeeds if a value of type A can be decoded at runtime */
+    (value: A): Validation<A>
+    /** succeeds if a value of type S can be decoded to a value of type A */
+    (s: S, context: Context): Validation<A>
+  }
   constructor(
     /** a unique name for this runtime type */
     readonly name: string,
     /** a custom type guard */
     readonly is: Is<A>,
     /** succeeds if a value of type S can be decoded to a value of type A */
-    readonly validate: Validate<S, A>,
+    protected readonly _validate: Validate<S, A>,
     /** converts a value of type A to a value of type S */
     readonly serialize: Serialize<S, A>
   ) {
-    this.maybe = (value: any) => validate(value, [{ key: '', type: this }])
+    this.validate = (s: any, c: any[] | null = null) => _validate(s, c || [{ key: '', type: this }])
   }
   pipe<B>(ab: Type<A, B>, name?: string): Type<S, B> {
     return new Type(
@@ -263,7 +268,7 @@ export class RefinementType<RT extends Type<any, any>, S, A> extends Type<S, A> 
   constructor(
     name: string,
     is: RefinementType<RT, S, A>['is'],
-    validate: RefinementType<RT, S, A>['validate'],
+    validate: RefinementType<RT, S, A>['_validate'],
     serialize: RefinementType<RT, S, A>['serialize'],
     readonly type: RT,
     readonly predicate: Predicate<A>
@@ -297,7 +302,7 @@ export class LiteralType<V extends string | number | boolean> extends Type<mixed
   constructor(
     name: string,
     is: LiteralType<V>['is'],
-    validate: LiteralType<V>['validate'],
+    validate: LiteralType<V>['_validate'],
     serialize: LiteralType<V>['serialize'],
     readonly value: V
   ) {
@@ -322,7 +327,7 @@ export class KeyofType<D extends { [key: string]: any }> extends Type<mixed, key
   constructor(
     name: string,
     is: KeyofType<D>['is'],
-    validate: KeyofType<D>['validate'],
+    validate: KeyofType<D>['_validate'],
     serialize: KeyofType<D>['serialize'],
     readonly keys: D
   ) {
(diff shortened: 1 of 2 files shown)


gcanti:
Thanks for this PR but I don't understand the meaning of following signature

```ts
(value: A): Validation<A>
```

If you can prove to the type checker that `value` has type `A` then there's no need for validations at all

mmkal:
At the boundaries of the application, it's very common to get some data from an external service or library, which we can't make type guarantees about, even if we give them types with typescript.
e.g.
```js
const name = await nameService.fromId(123)
const age = await ageService.fromId(123)
return Person.validate({ name, age })
```
What it name service sends back null even though it's not in the type definition? Say, the type definition is wrong because it's from definitelytyped and isn't maintained, or was written without strictNullChecks, or there's just a bug somewhere. For me the value is io-ts is that I can write code assuming I can trust the type system, because io-ts will verify it at runtime. Without having that type going into `validate` I can easily make silly mistakes that could have been caught at compile time:

```js
return t.validate({ fullName: name, age })
```

Of course we could make an intermediate String type to check name before using it to construct a person, but why make it a necessity?

To flip your question on its head - since most types anybody will end up using ultimately inherit from `Type<P, mixed>`, in practice it's always `mixed` being passed into the `t.validate(..., Person)` or `Person.validate(..., ...)` methods. What's the value in that? In cases where you're parsing a string, you'll have an `any` type which you'll be allowed to pass in to anything. Or, you'll have a well-formed Person, in which case you'll have no problem passing it into the validator. The only use case I can see for actually needing the `mixed` type in `validate` is when you know the type system is lying to you and want to pass in something that typescript thinks is invalid. Even then, it's very easy to workaround with casting:
```
const age: string = someLibrary.getAge(123)

const alice = Person.validate({ name: "alice", age }) // typescript error, age should be a number
const bob = Person.validate({ name: "bob", age } as any) // no tsc error, runtime check will still occur
```
There's even an example of this in one of the tests I wrote.

So I'd actually argue that all `validate` signatures should work this way, but it'd technically be a breaking change (but if you think about it, in almost all cases it'll only break code that's already wrong).

If you agree I'd be happy to see if the existing validate could be adapted! What do you think?

gcanti:
> Of course we could make an intermediate String type to check name before using it to construct a person, but why make it a necessity?

Because is the right thing to do. 

If you don't trust an API you should type its result with `mixed` and then validate its actual result as soon as possible, i.e at the boundaries of the system

```ts
const name = await nameService.fromId(123) // <= this is the boundary so validations go here
const age = await ageService.fromId(123)
const myname: string = name.trim() // what if you use name before validating?
return Person.validate({ name: myname, age })
```

mmkal:
You are right, it'll usually be better to validate as early as possible. But this feature wouldn't discourage that, it just provides the option to use an untrusted value directly (of course if you're trimming you can't escape validating before that no matter what):
```js
const p = Person.validate({ name: await nameService.fromId(123) })
```
This will give a useful error message: something like "expected string for property 'name', got null" which is nice. 

And what about type hints from the ide and my `fullName` bug example above?

Also, can you think of a use case where using `A`/ the `InterfaceOf` type would be a problem? Genuine  question as I am much less familiar with the usages of the library. The only example I've seen of code that would be broken by this is https://github.com/gcanti/io-ts/blob/f18deaa7c890717efe7370b42ecaf4c2fe6207ff/test/interface.ts#L35-L37 but those are *invalid* types being passed in. I want to break that code!

gcanti:
@mmkal just look at the signature

```ts
(value: A): Validation<A>
```

It clearly doesn't make much sense from the POV of a type system.

> it just provides the option to **use an untrusted value** directly

Actually that's exactly what I want to make impossible with io-ts

> of course if you're trimming you can't escape validating before that no matter what

I don't want to remember what I can or can't do, that's the type checker's duty

> Also, can you think of a use case where using A/ the InterfaceOf type would be a problem? 

If you type the untrusted API properly (i.e with `mixed`), your overloading doesn't type check anymore

```ts
declare const untrustedNameService: {
  fromId: (id: number) => t.mixed // instead of string
}
declare const trustedAgeService: {
  fromId: (id: number) => number
}

const name = untrustedNameService.fromId(123)
const age = trustedAgeService.fromId(123)
Person.validate({ name, age }) // type error: Type 'mixed' is not assignable to type 'string'
```

> And what about type hints from the ide and my fullName bug example above?

If you are validating the parts, there's no need to validate the whole (i.e. you can just `map`)

```ts
type Person = {
  name: string
  age: number
}

const getPerson = (id: number): Either<t.Errors, Person> =>
  t.validate(untrustedNameService.fromId(id), t.string).map(name => ({
    name,
    age: trustedAgeService.fromId(123)
  }))
```

PR #119, opened 2018-01-30
UPDATED - validate overload without target type parameter

mmkal:
@gcanti feedback taken on https://github.com/gcanti/io-ts/pull/118 - I have removed the `(value: A): Validation<A>` part, so now all this is is an overload for `MyType.validate(...)` which will pass in a sensible default context for you in the second parameter.

On reflection, it's cleaner this way - just a simple change to make a single parameter optional. Everything otherwise stays the same. Would you be happy to take this in? There's a usage example in the tests added.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -49,18 +49,23 @@ export class Type<S, A> implements Decoder<S, A>, Encoder<S, A> {
   readonly '_A': A
   // prettier-ignore
   readonly '_S': S
-  readonly maybe: (value: A) => Validation<A>
+  readonly validate: {
+    /** succeeds if a value of type A can be decoded at runtime */
+    (value: A): Validation<A>
+    /** succeeds if a value of type S can be decoded to a value of type A */
+    (s: S, context: Context): Validation<A>
+  }
   constructor(
     /** a unique name for this runtime type */
     readonly name: string,
     /** a custom type guard */
     readonly is: Is<A>,
     /** succeeds if a value of type S can be decoded to a value of type A */
-    readonly validate: Validate<S, A>,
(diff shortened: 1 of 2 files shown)


gcanti:
@mmkal `validate` and `serialize` are "old" names. I'm waiting for the next breaking change release to change them to `decode` and `encode`. 
So in the meanwhile and instead of an overloading, we could add a `decode` method without the context argument and deprecate `t.validate`, something along the lines of

```diff
@@ -31,6 +31,7 @@ export type InputOf<RT extends Type<any, any>> = RT['_S']
 export interface Decoder<S, A> {
   readonly name: string
   readonly validate: Validate<S, A>
+  readonly decode: (s: S) => Validation<A>
 }
 
 export interface Encoder<S, A> {
@@ -75,6 +76,10 @@ export class Type<S, A> implements Decoder<S, A>, Encoder<S, A> {
   asEncoder(): Encoder<S, A> {
     return this
   }
+  /** succeeds if a value of type S can be decoded to a value of type A */
+  decode(s: S): Validation<A> {
+    return validate(s, this)
+  }
 }
 
 export const identity = <A>(a: A): A => a
@@ -105,6 +110,7 @@ export const failure = <T>(value: mixed, context: Context): Validation<T> =>
 
 export const success = <T>(value: T): Validation<T> => new Right<Errors, T>(value)
 
+/** @deprecated use `type.decode(value)` instead */
 export const validate = <S, A>(value: S, type: Decoder<S, A>): Validation<A> =>
   type.validate(value, getDefaultContext(type))
```

mmkal:
closing since this should be covered by https://github.com/gcanti/io-ts/pull/123

PR #131, opened 2018-02-15
add strictInterfaceWithOptionals combinator to tests

gcanti:
..to prevent regressions. /cc @gunzip

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "lint": "tslint src/**/*.ts test/**/*.ts",
     "typings-checker":
       "typings-checker --allow-expect-error --project typings-checker/tsconfig.json typings-checker/index.ts",
-    "mocha": "TS_NODE_CACHE=false mocha -r ts-node/register test/*.ts",
+    "mocha": "TS_NODE_CACHE=false TS_NODE_PROJECT=test/tsconfig.json mocha -r ts-node/register test/*.ts",
     "prettier":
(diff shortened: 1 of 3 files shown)


gunzip:
a question, why not

```
): t.Type<
  t.TypeOfProps<R> & t.TypeOfPartialProps<O>,
  t.OutputOfProps<R> & t.OutputOfPartialProps<O>
> {

```

?

gcanti:
because the extracted type would be less readable

```ts
const T = strictInterfaceWithOptionals({ foo: t.string }, { bar: t.string }, 'T')

type TT = t.TypeOf<typeof T>
/*
type TT = {
    foo: string;
} & {
    bar?: string | undefined;
}
*/
```

versus

```ts
/*
type TT = t.TypeOfProps<{
    foo: t.StringType;
}> & t.TypeOfPartialProps<{
    bar: t.StringType;
}>
*/
```

gcanti:
@gunzip I changed my mind, I think

```ts
t.TypeOfProps<R> & t.TypeOfPartialProps<O>,
t.OutputOfProps<R> & t.OutputOfPartialProps<O>
```

is better (and more future proof, see https://github.com/gcanti/io-ts/issues/140)

gunzip:
ahh nice ! this is our implementation by now:
https://github.com/teamdigitale/digital-citizenship-functions/blob/master/lib/utils/types.ts#L107

PR #138, opened 2018-02-23
Internal: add push-slice-pop optimization

gcanti:
@sledorze the gist is

- `context` is _mutable_ (I feel a bit dirty for this) and I push / pop `ContextEntry`s while traversing the tree
- if an error is encountered then I take a snapshot of the current `context` by calling `context.slice()` in the `failure` API

Analysis:

In the best case (which we should optimize for) only one array is created and no snapshots at all.
In the worst case (all leafs are invalid) I guess it does less work as well

Benchmarks:

Baseline (from last PR)

```
space-object (good) x 666,376 ops/sec ±0.40% (88 runs sampled)
space-object (bad) x 587,146 ops/sec ±0.90% (85 runs sampled)
```

after this change

```
space-object (good) x 1,021,748 ops/sec ±0.39% (89 runs sampled)
space-object (bad) x 748,862 ops/sec ±0.64% (89 runs sampled)
```

The main point is: this change looks backward compatible if the `failure` API is consistently used, or am I wrong?

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -112,10 +112,15 @@ export const appendContext = (c: Context, key: string, type: Decoder<any, any>):
   return r
 }
 
+const pushContext = (c: Context, key: string, type: Decoder<any, any>): Context => {
+  c.push({ key, type })
+  return c
+}
+
 export const failures = <T>(errors: Errors): Validation<T> => new Left(errors)
 
 export const failure = <T>(value: mixed, context: Context): Validation<T> =>
-  failures([getValidationError(value, context)])
+  failures([getValidationError(value, context.slice())])
 
 export const success = <T>(value: T): Validation<T> => new Right<Errors, T>(value)
 
@@ -435,7 +440,8 @@ export const array = <RT extends Mixed>(
         const errors: Errors = []
         for (let i = 0; i < len; i++) {
           const x = xs[i]
-          const validation = type.validate(x, appendContext(c, String(i), type))
+          const validation = type.validate(x, pushContext(c, String(i), type))
+          c.pop()
           if (validation.isLeft()) {
             pushAll(errors, validation.value)
           } else {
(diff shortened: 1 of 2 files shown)


sledorze:
@gcanti I think it's a breaking change as the reporting will immediately break for end user that have implemented custom Types with the validate API (public).

---

Also, if a 'no breaking change' is no more a constraint; what would we choose?

From my subjective perspective we have two solutions that look similar in speed gain but:
- One is mutable (brittle) and the user should know it.
- The other has a different report structure, (which I see as a feature - having more information; the hierarchy) but is a little more subtil in the parameters to pass to the validate function (the additional  key).

I'm wondering if we should not try decide first of a solution (with enhancement or maybe another alternative) before implementing one in a hurry and change it at the next breaking change (soon).

---

I am saying that as 2.8 is around the corner and Yes, I would love to have a nicer solution than intersecting with partials to have optional members.. :)

gcanti:
> I think it's a breaking change as the reporting will immediately break for end user that have implemented custom Types with the validate API

You mean if the end user makes use of something else to create a failure other than the `failure` API right?

```ts
const MyString = new t.Type<string>(
  'string',
  (m): m is string => typeof m === 'string',
  // refuse to use the `failure` API here
  (m, c) => (typeof m === 'string' ? t.success(m) : t.failures([t.getValidationError(m, c)])),
  t.identity
)

const Person = t.type(
  {
    name: MyString,
    age: t.number
  },
  'Person'
)

console.log(PathReporter.report(Person.decode({})))
/*
[ 'Invalid value undefined supplied to : Person', // <= wrong error message here
  'Invalid value undefined supplied to : Person/age: number' ]
*/
```

That's my main concern too, if the `failure` API is always used then you won't notice a different behaviour. 

However I can't be sure of that in user land, so this change would be too dangerous.

Closing this, thanks for reviewing (and for the great work you did in the last few days on this subject).

sledorze:
@gcanti ah.. that's a good question; it may break 'working' but wrong code, so I think yes, we should release as a bug fix.

PR #164, opened 2018-04-25
Fix sample typo

NN---:


gcanti:
Thanks @NN---

NN---:
@gcanti I spent about half hour before understood that the problem is so simple :)

Issue #157, opened 2018-04-17
"t.intersection" fails validation when composed of "t.strict"s


FrancescoCioria:
## description
If I define a `t.intersection` of two `strict`s, when I try to validate a valid object with `.decode`, it fails.

It works correctly if I use `t.type` instead of `t.strict`

## how to reproduce
This works:
```ts
const A = t.type({
  a: t.string
});

const B = t.type({
  b: t.string
});

const C = t.intersection([A, B]);

C.decode({ a: 'a', b: 'b' }).isLeft(); // => false
```

This returns an error:
```ts
const A = t.strict({
  a: t.string
});

const B = t.strict({
  b: t.string
});

const C = t.intersection([A, B]);

C.decode({ a: 'a', b: 'b' }).isLeft(); // => true
```

Returned error:
```
Invalid value "b" supplied to : (StrictType<{ a: string }> & StrictType<{ b: string }>)/b: never
Invalid value "a" supplied to : (StrictType<{ a: string }> & StrictType<{ b: string }>)/a: never
```

gcanti:
When defined with `strict`, `A` and `B` are disjoint thus their intersection is empty

PR #166, opened 2018-05-02
New Feature: add `exact` combinator

gcanti:
`strict` is not flexible enough, this PR adds an `exact` combinator which is able to

> Is there a way to say "I want an object containing, optionally, any of these fields, but no other fields"? If you pass known optional fields, they are validated, and if you pass unknown fields they fail? t.strict requires fields to exist, and t.partial doesn't validate unknown

(https://twitter.com/nkariniemi/status/991396518082043905)

/cc @hoodunit

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,18 +1,19 @@
 {
   "name": "io-ts",
-  "version": "1.1.0",
+  "version": "1.1.1",
   "description": "TypeScript compatible runtime type system for IO validation",
-  "files": [
-    "lib"
-  ],
+  "files": ["lib"],
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
   "scripts": {
     "lint": "tslint -p tsconfig.json src/**/*.ts test/**/*.ts",
-    "typings-checker": "typings-checker --allow-expect-error --project typings-checker/tsconfig.json typings-checker/index.ts",
+    "typings-checker":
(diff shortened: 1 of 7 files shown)


gcanti:
Some examples

Works with `type` (same as `strict`, which we could deprecate)

```ts
const A = t.exact(t.type({ a: t.string }))

console.log(A.decode({ a: 'a' }).isRight()) // true
console.log(A.decode({ a: 'a', b: 1 }).isRight()) // false
```

Works with `partial`

```ts
const B = t.exact(t.partial({ b: t.string }))

console.log(B.decode({}).isRight()) // true
console.log(B.decode({ b: 'b' }).isRight()) // true
console.log(B.decode({ c: 1 }).isRight()) // false
```

Works with `intersection`

```ts
const C = t.exact(t.intersection([t.type({ a: t.string }), t.partial({ b: t.string })]))

console.log(C.decode({ a: 'a', b: 'b' }).isRight()) // true
console.log(C.decode({ a: 'a', c: 1 }).isRight()) // false
```

hoodunit:
This is nice. I tested it on our code base and it does exactly what I need. Thanks!

gcanti:
@FrancescoCioria this new combinator would also solve your use case here https://github.com/gcanti/io-ts/issues/157

@gunzip not sure if you are still using `strictInterfaceWithOptionals` but with this new combinator it would be implemented as

```ts
t.exact(t.intersection([t.interface(requiredProps), t.partial(optionalProps)])
```

Issue #190, opened 2018-07-18
Generic Types


peterharding:
I've been trying to represent this structure in `io-ts`:
```
export interface Response<T> {
  result: T[];
  _links: Links;
}

export interface Links {
  previous?: string;
  self: string;
  next?: string;
}
```

But after reading the README and trying a few things, I couldn't find a solution.
If there were a solution, I would imagine it to be something like:
```
const Self = iots.interface({
  self: iots.string,
});

const PreviousNext = iots.partial({
  previous: iots.string,
  next: iots.string,
});

export const Links = iots.intersection([Self, PreviousNext]);

const GenericResult = iots.interface<T>({
  result: iots.array(T),
});

export const Response = iots.intersection([GenericResult, Links]);
```
But the `interface<T>` syntax is not understood. Also, unless I misunderstand, recursive types will not solve this problem since they require a static type as a hint (which I don't have for a generic type). 

So, are generic types unsupported, or is there something I'm missing?

gcanti:
You can represent polimorphic runtime types with functions. For example `Array<T>` is represented by (the built-in)

```ts
iots.array: <RT extends Type<any, any, mixed>>(type: RT, name?: string) => ArrayType<RT, RT["_A"][], RT["_O"][], mixed>
```

So in your case

```ts
export const response = <RT extends iots.Mixed>(type: RT) =>
  iots.interface({
    result: iots.array(type),
    _links: Links
  })
```

peterharding:
Thanks @gcanti !
Perhaps this would be a good example to add to the README?

gcanti:
Sure, would you like to send a PR?

peterharding:
Sure! PR created here: https://github.com/gcanti/io-ts/pull/191

peterharding:
Resolved in https://github.com/gcanti/io-ts/pull/191, closing

PR #191, opened 2018-07-19
Update README to include Generic Types example

peterharding:
For https://github.com/gcanti/io-ts/issues/190

See: https://github.com/peterharding/io-ts/tree/issue-190-generic-readme-example#generic-types

screen shot 2018-07-20 at 10 36 40 am

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,9 +369,9 @@ Would be:
 ```ts
 import * as t from 'io-ts';
 
-const ResponseBody = <RT extends iots.Mixed>(type: RT) =>
-  iots.interface({
-    result: iots.array(type),
+const ResponseBody = <RT extends t.Mixed>(type: RT) =>
+  t.interface({
+    result: type,
     _links: Links
   });
 
@@ -383,7 +383,7 @@ const Links = t.interface({
 And used like:
 ```ts
 const UserModel = t.interface({ name: t.string });
-functionThatRequiresRuntimeType(ResponseBody(UserModel), ...params);
+functionThatRequiresRuntimeType(ResponseBody(t.array(UserModel)), ...params);
 ```
 
 # Tips and Tricks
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -352,6 +352,40 @@ DateFromString.decode('foo')
...
+  next: string;
+}
+```
+Would be:
+```ts
+import * as t from 'io-ts';
+
+const ResponseBody = <RT extends iots.Mixed>(type: RT) =>
+  iots.interface({
+    result: iots.array(type),
```
peterharding: Should probably pass in the array like `ResponseBody(t.array(UserModel))` instead of assuming it's an array here.

README.md
```diff
@@ -352,6 +352,40 @@ DateFromString.decode('foo')
...
+interface Links {
+  previous: string;
+  next: string;
+}
+```
+Would be:
+```ts
+import * as t from 'io-ts';
+
+const ResponseBody = <RT extends iots.Mixed>(type: RT) =>
```
peterharding: Should be `t` instead of `iots` to match the rest of the README

README.md
gcanti: If you are representing

```ts
interface ResponseBody<T> {
  result: T[];
  _links: Links;
}
```

I think that the previous version was correct. The current implementation

```ts
const ResponseBody = <RT extends t.Mixed>(type: RT) =>
  t.interface({
    result: type,
    _links: Links
  });
```

is representing

```ts
interface ResponseBody<T> {
  result: T;
  _links: Links;
}
```

instead

peterharding: yeah I think it's better to use
```
interface ResponseBody<T> {
  result: T;
  _links: Links;
}
```
but I forgot to update the original interface, will do that!

gcanti:
Thanks @peterharding

Issue #195, opened 2018-07-31
Unexpected validation path from partial type


mwikstrom:
Hi. Thanks for providing this lib.

I'm seeing unexpected validation paths when using partial types.

To me it looks like an extra numeric path segment is added to all partial properties, like `a/b/0` instead of just `a/b`.

Please see the following test that reproduce the problem:

~~~~typescript
import * as t from "io-ts";

test("io-ts partial validation", () => {
    // Define a partial type
    const partialType = t.partial({
        a: t.number,
    });

    // Bad input to decode
    const badInput = { a: null };

    // Get validation result
    const result = partialType.decode(badInput);

    // We know there is a single error. Extract its path.
    const path = result.mapLeft(
        errors => errors[0].context.map(context => context.key).filter(key => !!key).join("/"),
    ).value;

    // Path should be `a`. But expectation fails. Actual value is `a/0`
    expect(path).toBe("a"); 
});
~~~~

Is this a bug or is there some reasoning behind that extra path segment?

(I'm using io-ts version 1.2.1)

mwikstrom:
After looking at the implementation I see that partials are described by using unions.

[Line 600](https://github.com/gcanti/io-ts/blob/master/src/index.ts#L600):
~~~~typescript
partials[keys[i]] = union([types[i], undefinedType])
~~~~

So, the problem (if it is one) stems from the fact that the union type appends an integer key to the validation context for each type in the union.

[Line 734](https://github.com/gcanti/io-ts/blob/master/src/index.ts#L734):
~~~~typescript
const validation = type.validate(m, appendContext(c, String(i), type))
~~~~

What's the reason for appending that integer key? To me it looks wierd, but I've just started using this lib, so I might be missing something important.

sledorze:
It is a (correct) feature; the error reported incorporate the failing union entry (here 0), otherwise you would have no clue which entry fails.

mwikstrom:
Ok.

But union entry index (0 or 1) in a partial/optional property path does not make any sense. 

So maybe partials should be implemented in another way?

Maybe by not using unions, or maybe by stripping out that extra path from validation errors?

sledorze:
Maybe by adding a specific solution for partials.
I'm not sure if its worth it.

mwikstrom:
I've added PR #198 that do not use unions in partials. What's your opinion on that?

PR #198, opened 2018-08-15
add optional type and use it in partials

mwikstrom:
The purpose of this change is to *not* expose "union type index" in validation errors when using partials (issue #195).

This is accomplished by adding a new optional type, which works like a union between a specified type and the undefined type, and using it instead of unions inside partials.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,7 @@ import * as t from 'io-ts'
 | function                  | `Function`                              | `t.Function`                                          |
 | literal                   | `'s'`                                   | `t.literal('s')`                                      |
 | partial                   | `Partial<{ name: string }>`             | `t.partial({ name: t.string })`                       |
+| optional                  | `A | undefined`                         | `t.optional(A)`                                       |
 | readonly                  | `Readonly<T>`                           | `t.readonly(T)`                                       |
 | readonly array            | `ReadonlyArray<number>`                 | `t.readonlyArray(t.number)`                           |
 | type alias                | `type A = { name: string }`             | `t.type({ name: t.string })`                          |
--- a/README.md
+++ b/README.md
@@ -205,7 +205,7 @@ import * as t from 'io-ts'
 | function                  | `Function`                              | `t.Function`                                          |
 | literal                   | `'s'`                                   | `t.literal('s')`                                      |
 | partial                   | `Partial<{ name: string }>`             | `t.partial({ name: t.string })`                       |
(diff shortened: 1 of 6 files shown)


sledorze:
@mwikstrom bird view that's appear to be better on error messages (and maybe performance too).

The only cons I see is that it adds a new combinator to this `core` lib, it could have go into `io-ts-types` but partial partial would still suffer from the same `down-side`.

My personal stance is to go for it (after a deeper review).

gcanti:
@mwikstrom thanks for this PR. In TypeScript when you define an optional property its value has type `T | undefined`

```ts
type Foo = {
  bar?: string
}
/*
type Foo = {
    bar?: string | undefined;
}
*/
```

so the current implementation, which uses a union, looks correct. 

Now having an alias/combinator for `T | undefined` (this PR) is nice, however using it in `partial` would be a breaking change.

sledorze:
@gcanti oh.. you mean we loose the '?' ?
my bad I've totally missed that @mwikstrom ..

gcanti:
@sledorze no, from what I can understand @mwikstrom is proposing to use `optional` in the `partial` implementation, so the `?`s would be still there. However changing how the errors are built is a breaking change. 

Also, exposing an `optional` combinator can be confusing for users, they could end up using that instead of `partial`.

mwikstrom:
@gcanti in my opinion it is incorrect that partial types include union type index in validation errors.

It is an implementation detail that unions are used inside partials. Therefore, exposing union type index is (again in my opinion) confusing and it does not really serve any purpose.

Therefore, I'd say this is a fix rather than a breaking change.

Care to explain your thoughts on this?

gcanti:
@mwikstrom this syntax

```ts
type Foo = {
  bar?: string
}
```

means 2 things

- `bar`'s value has type `string | undefined` (i.e., a union, as implemented in `partial`)
- the `bar` property may not be there at all

sledorze:
I agree with @mwikstrom about a cleaner error report
and I acknowledge it's a breaking change, requiring a major version or be exposed via another combinator.

Maybe `io-ts-types` is a better place then?

rzane:
Is anything going to change in response to this PR? It would be really nice to avoid having the union type index.

For my use case, I'm building a CLI tool that has a config file that a user creates. I'd like to validate the config file to assist the user. This library seems like the perfect way to do this, but, I worry that the union type index is going to confuse users of my command line tool.

Is there a different solution to this problem?

PR #199, opened 2018-08-19
Use NonEmptyArray for representing validation errors

gabro:
As far as I can tell, if validation errors are present they should not be empty.

This PR replaces `Array<ValidationError>` with `NonEmptyArray<ValidationError>` whenever it makes sense.

Note that so far I've kept the `PathReporter` to return `Array<string>` because changing that would've been a breaking change in the APi.

I think it makes sense to do it, since as a user I would expect to be able to do stuff like `.head` on the errors, but I wanted to ask for feedback first.

### DIFF
--- a/src/PathReporter.ts
+++ b/src/PathReporter.ts
@@ -1,5 +1,6 @@
 import { Reporter } from './Reporter'
 import { Context, getFunctionName, ValidationError } from './index'
+import { NonEmptyArray, nonEmptyArray } from 'fp-ts/lib/NonEmptyArray'
 
 function stringify(v: any): string {
   return typeof v === 'function' ? getFunctionName(v) : JSON.stringify(v)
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -738,7 +741,7 @@ export const union = <RTS extends Array<Mixed>>(
           pushAll(errors, validation.value)
         }
       }
-      return failures(errors)
+      return failures((fromArray(errors) as Some<Errors>).value)
```
gabro: the cast is due to the early `return` inside the for loop. If we got to this point the `errors` array is already known to be non-empty.

src/PathReporter.ts
```diff
@@ -13,14 +14,14 @@ function getMessage(v: any, context: Context): string {
...
 
-export function success(): Array<string> {
-  return ['No errors!']
+export function success(): NonEmptyArray<string> {
+  return nonEmptyArray.of('No errors!')
 }
 
 export const PathReporter: Reporter<Array<string>> = {
-  report: validation => validation.fold(failure, success)
+  report: validation => validation.fold(failure, success).toArray()
```
gabro: this is where I would love for this to be `Reporter<NonEmptyArray<string>>`, but it would be a breaking change, so let's talk about it.

gcanti:
@gabro thanks for this PR, I'm afraid I can't accept it though, it contains breaking changes (`Errors`, `failure` and `success` are public APIs),

For the record I plan to refactor the errors in `io-ts@2.0.0`, I'll open issue about it when the time comes

gabro:
👌 eager to see the new errors in 2.0 :)

PR #207, opened 2018-09-12
failing test for keyof

mmkal:
I'd expect keys to be a dictionary of the actual keys, rather than the dictionary passed in. Since there's not a test for this I'm not sure what the intended behaviour is.

Todo:
- [ ] Follow up with a 'fix' to `KeyOfType` to make `.keys` return what this test is expecting

### DIFF
--- a/dtslint/index.ts
+++ b/dtslint/index.ts
@@ -37,7 +37,7 @@ type Assert1 = t.TypeOf<typeof L1> // $ExpectType "a"
 
 const K1 = t.keyof({ a: true, b: true })
 type Assert2 = t.TypeOf<typeof K1> // $ExpectType "a" | "b"
-type Assert3 = typeof K1.keys // $ExpectType { a: "a", b: "b" }
+type KeyofKeys = typeof K1.keys // $ExpectType { a: "a", b: "b" }
 
 //
 // default types
--- a/dtslint/index.ts
+++ b/dtslint/index.ts
@@ -37,7 +37,7 @@ type Assert1 = t.TypeOf<typeof L1> // $ExpectType "a"
 
 const K1 = t.keyof({ a: true, b: true })
 type Assert2 = t.TypeOf<typeof K1> // $ExpectType "a" | "b"
-type KeyofKeys = typeof K1.keys // $ExpectType { a: "a", b: "b" }
+type KeyofKeys = typeof K1.keys // $ExpectType { a: "a"; b: "b"; }
 
 //
 // default types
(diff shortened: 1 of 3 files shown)


gcanti:
@mmkal Thanks for this PR, looks like a breaking change though.

> I'm not sure what the intended behaviour is

The intended behaviour is specified by https://github.com/gcanti/io-ts/blob/0201cdc02fd4331088e51d8a32343eff367b87b3/src/index.ts#L359

> I'd expect keys to be a dictionary of the actual keys, rather than the dictionary passed in

What's the benefit over the current implementation?

mmkal:
The main benefit is that a map of key -> key is more useful than the dummy object usually passed into `keyof`. The way I see keyof usually used is:
```typescript
const Direction = t.keyof({ up: 1, down: 1, left: 1, right: 1 })
```
being able to do this is pretty useful:
```typescript
type Direction = keyof typeof Direction.keys
const directions = Direction.keys

const moveThing = (d: Direction) => { ... }

moveThing(directions.up) // equivalent to `moveThing('up')` but avoids having to use a quoted string so has better intellisense etc.
```

I can't really think of anything I would use the original dictionary, i.e.`{ up: 1, down: 1, left: 1, right: 1 }` for, since the value of `up`, `down`,`left` and `right` are all just the dummy value 1.

gcanti:
> a map of key -> key is more useful than the dummy object

Ok but whether is dummy or not is up to you, let's say is imported from another module 

```ts
// countries.ts
export const countries = {
  IT: 'Italy',
  FR: 'France'
}

// ---

import { countries } from './countries'

const Country = t.keyof(countries)
```

and then what you do with `keys` is again up to you, `io-ts` shouldn't interfere with that.

Also you can always create a key -> key map on the fly if needed

```ts
declare function keyMirror<A extends PropertyKey>(keys: Array<A> | Record<A, unknown>): { [K in A]: K }

const directions = keyMirror(Direction.keys)
```

mmkal:
Good point. There is also a benefit in not having `keyof` and `KeyofType` be subtly different. That `keyMirror` thing is pretty much exactly what we currently do. It'd be nice not to have to but no big deal. Closing.

mmkal:
@gcanti related, and potentially would be a better solution: what about something like:
```typescript
const stringEnum = <T extends string>(values: T[], name?: string): t.KeyofType<{ [P in T]: P }> =>
  t.keyof(values.reduce((obj, next) => ({ ...obj, [next]: next }), {} as any), name)

export { stringEnum as enum }
```
used like
```typescript
const Direction = t.enum(['up', 'down', 'left', 'right'])
```
To avoid the need for a dummy object completely (which is annoying to type out, and  without introducing ambiguity/complexity into `keyof`.

gcanti:
Yeah, that's why my `keyMirror` above takes an array as well as an object

```ts
const Direction = t.keyof(keyMirror(['up', 'down', 'left', 'right']))
```

PR #213, opened 2018-10-02
DateType added

mlegenhausen:


gcanti:
@mlegenhausen what's the use case? Usually a Date will be decoded from a string or a number, this type seems for type checking rather than decoding.

mlegenhausen:
@gcanti you are right my usecase is type checking. `io-ts` seems perfect when your data source is `JSON` where the `Date` type gets encoded as a `string` or `number` but when working with other systems that are not `JSON` based you can end up in "mixed" environments where I want to handle `null` types with `createOptionFromNullable` but have actual `Date` objects returned from these systems. In my case I want to parse query results returned by `node-postgres` which does the `Date` conversion for me and I only want to validate that the returned property is really a `Date` object. That is why a want to add the `DateType` to `io-ts`.

gcanti:
@mlegenhausen I see, thanks for explaining. It would be ok for you to move this PR to io-ts-types? I'd like to keep the core of this library lean

mlegenhausen:
Sure

mlegenhausen:
PR: https://github.com/gcanti/io-ts-types/pull/65

PR #219, opened 2018-10-25
Bug Fix: dictionary type should not allow arrays, fix #218

### LINKED ISSUE #218: Runtime dictionary type allows arrays, derived static type does not
Version: 1.3.1

I'm a bit confused by the behavior of this library regarding arrays vs. dictionaries. 
I was expecting that a dictionary type would not accept arrays. However, I found that runtime io-ts dictionary types do allow arrays, whereas the derived static types do not.

```typescript
const Dictionary = t.dictionary(t.string, t.number);
const simpleArray = [1, 2, 3];
const test: t.TypeOf<typeof Dictionary> = simpleArray; // compiler complains
ThrowReporter.report(Dictionary.decode(simpleArray)); // does not throw
```

It seems that this is not described in the documentation or any (possibly closed) issues. Is this the intended behavior?

I did find a workaround for what I wanted to accomplish (using refinement):

```typescript
const RefinedDictionary = t.refinement(Dictionary, dictionary => !Array.isArray(dictionary));
const test2: t.TypeOf<typeof RefinedDictionary> = simpleArray; // compiler complains
ThrowReporter.report(RefinedDictionary.decode(simpleArray)); // throws
```

gcanti:
Unless the codomain is `any` (or `unknown`)

```ts
declare const as: Array<number>
declare const n: Number
declare const d: Date

const x1: Record<string, number> = as // error
const x2: Record<string, number> = n // error
const x3: Record<string, number> = d // error

const x4: { [key: string]: any } = as // ok
const x5: { [key: string]: any } = n // ok
const x6: { [key: string]: any } = d // ok
```

/cc @curledUpSheep

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "io-ts",
-  "version": "1.3.0",
+  "version": "1.3.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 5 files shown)


curledUpSheep:
Looks good to me :+1: 
I also tested the new code using some test cases of my own (for the runtime types I built on top of io-ts), the new code indeed removes the need for my workaround.

Good call on those cases with codomain `any `or `unknown`, had no idea that those behaved differently.

gcanti:
@curledUpSheep thanks for reviewing, I'll release a patch asap

PR #232, opened 2018-11-08
Test: `Type.prototype.pipe` should accept a type with a narrower outp…

### LINKED ISSUE #230: More reusable types with Pipe..
Currently Pipe signature is:

```typescript
pipe<B>(ab: Type<B, A, A>, name?: string): Type<B, O, I>;
```

This means the ab Type must exactly match `A` type.
I consider that if a Type can handle `unknown` has its input Type, then it should legitimately be used as `ab` type param.

The pipe signature would then become:

```typescript
pipe<B>(ab: Type<B, A, A> | Type<B, A, unknown>, name?: string): Type<B, O, I>;
```

@gcanti what do you think?

gcanti:
…ut, closes #230

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -55,9 +55,13 @@ export class Type<A, O = A, I = mixed> implements Decoder<I, A>, Encoder<A, O> {
     readonly encode: Encode<A, O>
   ) {}
 
-  pipe<B, IB, A extends IB, OB extends A>(this: Type<A, O, I>, ab: Type<B, OB, IB>, name?: string): Type<B, O, I> {
+  pipe<B, IB, A extends IB, OB extends A>(
+    this: Type<A, O, I>,
+    ab: Type<B, OB, IB>,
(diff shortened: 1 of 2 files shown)


mlegenhausen:
@gcanti can you provide some documentation how `pipe` is used?

gcanti:
@mlegenhausen maybe with an example?

# Piping

You can pipe two types if their type parameters do align

```ts
const NumberDecoder = new t.Type<number, string, string>(
  'NumberDecoder',
  t.number.is,
  (s, c) => {
    const n = parseFloat(s)
    return isNaN(n) ? t.failure(s, c) : t.success(n)
  },
  String
)

// NumberFromString: t.Type<number, string, t.mixed>
const NumberFromString = t.string.pipe(
  NumberDecoder,
  'NumberFromString'
)
```

mlegenhausen:
@gcanti examples are always good :wink:. Thanks 😄

PR #235, opened 2018-11-22
use Compact in Intersection signatures (fixes #234)

### LINKED ISSUE #234: Can't correctly constraint Types involving intersection based on their member output type
This should generate an error due to `b` being a number

```typescript
const q : t.Type<{ a: string, b: number }, { [k: string]: string  } > = t.intersection([
    t.interface({
      a: t.string
    }),
    t.interface({
      b: t.number
    })
  ])
```

sledorze:


gcanti:
Thanks @sledorze 👍

Issue #140, opened 2018-02-26
[RFC] TypeScript 2.8 and optional fields


gcanti:
from https://github.com/gcanti/io-ts/pull/138#issuecomment-368131968

> as 2.8 is around the corner, I would love to have a nicer solution than intersecting with partials to have optional members

There's a branch (`optional`) with a proposal that makes use of conditional types. The gist is

- added an `OptionalType` and `optional` combinator
- added `RequiredKeys` and `OptionalKeys` types (<= using conditional types)
- changed `TypeOfProps` and `OutputOfProps` accordingly

Result

```ts
const Person = t.type({
  name: t.string,
  age: t.optional(t.number) // new optional combinator
})

type T = t.TypeOf<typeof Person>
/*
type T = {
    name: string;
} & {
    age?: number | undefined;
}
*/
```

Bonus point: is backward compatible (AFAIK)

/cc @sledorze

EDIT

Note that this is different from make a union with `t.undefined`

```ts
const Person = t.type({
  name: t.string,
  age: t.optional(t.number),
  foo: t.union([t.number, t.undefined])
})

type T = t.TypeOf<typeof Person>
/*
type T = {
    name: string;
    foo: number | undefined; // foo can be undefined but the key is required
} & {
    age?: number | undefined;
}
*/

const x: T = { name: 'Giulio' } // Property 'foo' is missing in type '{ name: string; }'.
```

gcanti:
Also /cc @giogonzo, @mattiamanzati (mobx-state-tree), @pelotom (runtypes)

giogonzo:
nice! This code can be improved using the new combinator: https://github.com/gcanti/io-ts-codegen/blob/master/src/index.ts#L491-L505

sledorze:
@gcanti is there a comparison speed wise between the new and historic way to encode an object via intersection and partial?

gcanti:
@sledorze just added a perf suite, looks good

```ts
// optional
const T1 = t.type({
  a: t.string,
  b: t.optional(t.number)
})

// intersection
const T2 = t.intersection([t.type({ a: t.string }), t.partial({ b: t.number })])
```

Results

```
optional (valid) x 6,283,783 ops/sec ±0.77% (86 runs sampled)
intersection (valid) x 4,483,965 ops/sec ±0.75% (88 runs sampled)
optional (invalid) x 4,671,694 ops/sec ±0.91% (88 runs sampled)
[...]

Issue #196, opened 2018-08-08
Make it easier to specify partial types


tonyxiao:
There's this amazing tool to generate io-ts from json - https://transform.now.sh/json-to-io-ts

However it doesn't work with optional attributes. The current way (https://github.com/gcanti/io-ts#mixing-required-and-optional-props) of specifying partial is really complex / cumbersome. Could we support syntax like

```ts
const MyType = t.type({
  requiredField: t.number,
  optionalField: t.optional(t.string)
})
```

gcanti:
I wish... (see https://github.com/gcanti/io-ts/issues/140)

tonyxiao:
awww :( That's too bad. Any workaround you would recommend?

Issue #83, opened 2017-11-01
Embedded optional types?


jez9999:
The documentation suggests making optional members of a type using intersection:

```
const MandatoryRunSettings = t.interface({
	appName: FilledString,
});

const OptionalRunSettings = t.partial({
	shortAppName: FilledString,
});

const RunSettings = t.intersection([MandatoryRunSettings, OptionalRunSettings]);
```

However this can start to look a bit ugly if you have lots of members of a type and want related members to be next to one another.  Could we implement an "optional" tag for types so that we can just associate that data with the type?  Then we could say something like:

```
const RunSettings = t.interface({
	appName: FilledString,
	shortAppName: t.OptionalType<FilledString>,
});
```

gcanti:
AFAIK is not possible at the field level, see here for an explanation https://github.com/gcanti/io-ts/issues/56

jez9999:
I posted a comment in #56, please respond.  I still don't see what the problem is with this.

Issue #56, opened 2017-07-06
Simpler optional props syntax?


mindjuice:
I'm using `io-ts` in a current project and its working quite well, but there is one thing I'm not so keen on.

The current mechanism for creating an interface with some optional props is rather verbose.

```typescript
const A = t.interface({
  foo: t.string
})

const B = t.partial({
  bar: t.number
})

const C = t.intersection([A, B])

type CT = t.TypeOf<typeof C>
```

Would it be possible to have something like the following instead, even if, under the hood, it ended up creating the more complex version?

```typescript
const A = t.interface({
  foo: t.string,
  bar: t.optional(t.number)
})

type CT = t.TypeOf<typeof C>
```

I haven't yet looked at the code, but I would think you could iterate over the props and collect the ones set to optional and create a `t.interface()` and a `t.partial()` and combine them with `t.intersection()` as above.

typeetfunc:
@mindjuice https://github.com/pelotom/runtypes/issues/15#issuecomment-307601297

mindjuice:
Sorry, I'm not sure what bit of that conversation you want to call out.  It sounds like the same basic question on a different project.  Are you referring to the part where they say it's not possible in TypeScript?

I'm not asking to do anything different in the end than what the long-winded version already does, so it seems possible from a types point of view.

Or am I missing something in the referenced thread?

sledorze:
@mindjuice indeed, not possible.
To understand why, just have a look at how it's implemented.

gcanti:
@mindjuice the problem is that making a prop optional at the type level means to use a mapped type and the `?` modifier *on the keys*

```ts
export type PartialOf<P extends Props> = { [K in keyof P]?: TypeOf<P[K]> }
```

so I always need at least *two* mapped types, one for the required props and one for the optional props. 

Some boilerplate can be avoided by defining a helper though

```ts
export function mixed<R extends t.Props, O extends t.Props>(
  required: R,
  optional: O,
  name?: string
): t.IntersectionType<[t.InterfaceType<R>, t.PartialType<O>], t.InterfaceOf<R> & t.PartialOf<O>> {
  return t.intersection([t.interface(required), t.partial(optional)], name)
}

const C = mixed(
  {
    foo: t.string
  },
  {
    bar: t.number
  }
)

type Clean<T> = Pick<T, keyof T>

type CT = Clean<t.TypeOf<typeof C>>
/*
type CT = {
    foo: string;
    bar?: number | undefined;
}
[...]

PR #266, opened 2019-01-23
[POC] partial partial

mmkal:
@gcanti @sledorze 

Addresses https://github.com/gcanti/io-ts/issues/140 (I think).

I'm not totally sure I understand what the issue mentioned in https://github.com/gcanti/io-ts/issues/140#issuecomment-385632344 is, I tried pasting that code into `dtslint/index.ts` and it didn't seem to have cause problems I could see.

I also don't know if this implementation is the same as the POC from the `optional` branch mentioned in the OP because I can't find the branch (I guess it's been deleted).

I implemented it as a separate helper, `partialPartial` rather than updating the `type` export, just for the sake of a proof-of-concept. @gcanti @sledorze if this has the same issue as the previous POC, could you give a code sample that will fail dtslint?

### DIFF
--- a/dtslint/index.ts
+++ b/dtslint/index.ts
@@ -587,3 +587,14 @@ declare function withValidation<L, A>(
 declare const fa: TaskEither<string, void>
 
 withValidation(t.void, () => 'validation error', fa)
+
+const Person = t.partialPartial({
+  name: t.string,
+  age: t.optional(t.number),
+})
+type Person = t.TypeOf<typeof Person>
+
+const bob: Person = { name: 'Bob' }
+const bob30: Person = { name: 'Bob', age: 30 }
+// $ExpectError
+const nobody: Person = {}
--- a/src/index.ts
+++ b/src/index.ts
@@ -820,13 +820,53 @@ export interface Props {
  * @since 1.5.3
  */
 export interface TypeC<P extends Props>
-  extends InterfaceType<P, { [K in keyof P]: TypeOf<P[K]> }, { [K in keyof P]: OutputOf<P[K]> }, unknown> {}
+extends InterfaceType<P, { [K in keyof P]: TypeOf<P[K]> }, { [K in keyof P]: OutputOf<P[K]> }, unknown> {}
+
+interface Optional {
+  optional: true
+}
+
+export const optional = <T>(rt: Type<T>, name?: string) => {
+  const unionType = union([rt, nullType, undefinedType], name)
+  return Object.assign(unionType, { optional: true } as Optional)
+}
+
+type OptionalPropsKeys<P extends Props> = { [K in keyof P]: P[K] extends Optional ? K : never }[keyof P]
+type OptionalPropsTypes<P extends Props> = { [K in OptionalPropsKeys<P>]?: P[K] extends Optional ? P[K]['_A'] : never }
+type OptionalPropsOutputs<P extends Props> = {
+  [K in OptionalPropsKeys<P>]?: P[K] extends Optional ? P[K]['_O'] : never
+}
+type RequiredPropsKeys<P extends Props> = { [K in keyof P]: P[K] extends Optional ? never : K }[keyof P]
+type RequiredPropsTypes<P extends Props> = { [K in RequiredPropsKeys<P>]: P[K] extends Optional ? never : P[K]['_A'] }
+type RequiredPropsOutputs<P extends Props> = { [K in RequiredPropsKeys<P>]: P[K] extends Optional ? never : P[K]['_O'] }
+
+export const partialPartial = <P extends Props>(
+  props: P,
+  name?: string
+): Type<OptionalPropsTypes<P> & RequiredPropsTypes<P>, OptionalPropsOutputs<P> & RequiredPropsOutputs<P>, mixed> => {
+  let someOptional = false
+  const [optionalProps, requiredProps] = [{}, {}] as Props[]
+  for (const key of Object.keys(props)) {
+    const val: any = props[key]
+    if (val.optional) {
+      someOptional = true
+      optionalProps[key] = val
+    } else {
+      requiredProps[key] = val
+    }
+  }
+  if (someOptional) {
+    return intersection([type(requiredProps), partial(optionalProps)], name) as any
+  } else {
+    return type(props, name) as any
+  }
+}
 
 /**
  * @alias `interface`
  * @since 1.0.0
  */
-export const type =<P extends Props>(props: P, name: string = getNameFromProps(props)): TypeC<P> => {
+export const type = <P extends Props>(props: P, name: string = getNameFromProps(props)): TypeC<P> => {
   const keys = Object.keys(props)
   const types = keys.map(key => props[key])
   const len = keys.length
--- a/src/index.ts
+++ b/src/index.ts
@@ -826,7 +826,7 @@ export interface TypeC<P extends Props>
  * @alias `interface`
  * @since 1.0.0
  */
-export const type = <P extends Props>(props: P, name: string = getNameFromProps(props)): TypeC<P> => {
+export const type =<P extends Props>(props: P, name: string = getNameFromProps(props)): TypeC<P> => {
   const keys = Object.keys(props)
   const types = keys.map(key => props[key])
   const len = keys.length
--- a/test/partialPartial.ts
+++ b/test/partialPartial.ts
@@ -1,0 +1,18 @@
+import * as t from '../src/index'
+import { Right, Left } from 'fp-ts/lib/Either'
+
+describe('partialPartial', () => {
+  it('handles some optional props', () => {
+    const Person = t.partialPartial({ name: t.string, age: t.optional(t.number) })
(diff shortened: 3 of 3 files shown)


gcanti:
> I also don't know if this implementation is the same as the POC from the optional branch 

Yes, the gist was trying to use conditional types. Indeed I see the same problem

```ts
export const shrinked = <Type extends t.Type<A, O>, A = any, O = any>(
  type: Type
): t.Validation<{ t: 'v'; v: t.TypeOf<Type> }> => {
  const T = t.partialPartial({
    t: t.literal('v'),
    v: type
  })
  const res = T.validate(1 as any, [])
  res.map(x => {
    x.t // Ok
    x.v // Error: Property 'v' does not exist on type 'OptionalPropsTypes<{ t: LiteralC<"v">; v: Type; }> & RequiredPropsTypes<{ t: LiteralC<"v">; v: Type; }>'
  })
  return res // Errors as v is missing
}
```

mmkal:
@gcanti ah yes, I copy and pasted the example without changing `t.interface` to `t.partialPartial` 🤦‍♂️ 

Updated to fix for that problem, and add a test for it to dtslint to ensure no regressions. The only change needed to `shrinked` (aside from a more explicit name) was to make `v` in the return type optional, since it might be. However I also added an identical example with no return type (i.e. inferred), because the compiler is then smart enough to do the right thing - make `v` optional when an optional type is passed in, required otherwise. There's an `$ExpectError` covering this second scenario too.

mmkal:
@gcanti any thoughts on this? The issue mentioned in https://github.com/gcanti/io-ts/issues/140#issuecomment-385632344 is now addressed, and tested in https://github.com/gcanti/io-ts/pull/266/files#diff-a40ed4e571c91fc09060cdd4d991ff2aR609

gcanti:
@mmkal that's an interesting approach but I wouldn't touch the core `type` combinator, using an intersection is slightly more verbose but also seems more future proof

mmkal:
I agree, that's why I put it in a separate helper to make sure it doesn't interfere with existing code. It uses an intersection type as an implementation detail. Currently called `partialPartial` as a placeholder but there's probably a better name for it?

mmkal:
@gcanti it solves an issue that comes up a _lot_: https://github.com/gcanti/io-ts/issues/196 https://github.com/gcanti/io-ts/issues/140 https://github.com/gcanti/io-ts/issues/83 https://github.com/gcanti/io-ts/issues/56

You yourself said "I wish" in that latest one. And it seems at least with the latest and greatest typescript versions it's now possible. In the RFC issue you created @sledorze pointed out that it wouldn't work using abstract types passed into a function. In this PR I believe that scenario is handled successfully. Unless I've missed something (@sledorze - what do you think?) - now we know that it can be done, why not include it in the library since it's such a highly-requested feature, and significantly improves the syntax?

This PR is only a POC to figure out if the compiler can be persuaded to make the static type match the runtime type. The API design could be a follow-up step. This PR is really just the (potential) implementation detail.

Questions over whether it should it go into the `type` combinator, or be its own thing are a little above my pay-grade as I'm not a maintainer of this project.

I'd be happy to give my opinion (my slight preference would be that it does stay in the `type` combinator, so there isn't a separate one that needs to be used by).

VanTanev:
@gcanti @mmkal Maybe this can become a part of https://github.com/gcanti/io-ts-types?

mmkal:
It could. But having thought about it more since the original conversation and had a look at some of the other changes that have gone in, I definitely feel it'd be better as a first-class citizen. i.e. instead of a new combinator `partialPartial` the implementation of `type` should be updated. This could ensure that we'd still get expected behaviour with some of the other combinators like `union`, `intersection`, `strict`, etc. 

@gcanti would you be open to merging an update to this PR that did that? I'm happy to do the work but wouldn't want it to sit in review for a long time. The change would affect existing code, meaning merge conflicts would be very likely if it's waiting for too long.

gcanti:
> would you be open to merging an update to this PR that did that?

@mmkal thanks for this POC but no, I confirm my previous comments

mmkal:
I assume you're talking about the future-proof comment. That's the only one that applies to this suggestion. If so, are you saying you think the types could be broken by future typescript versions - I'd be interested to hear how/why. 

Do you at least agree it'd be better to enable optional properties in some form, as you've suggested in the past? And in that case, is the issue with this PR about the implementation/maintainability, or did you just change your mind (or is there something else)?

gcanti:
> are you saying you think the types could be broken by future typescript versions 

Yes it's possible (and it wouldn't be the first time IME) but I'm more concerned by the future directions of `io-ts` which I can't completely forsee now.

> Do you at least agree it'd be better to enable optional properties in some form, as you've suggested in the past?

Optional properties are possible from the beginning via `intersection` and `partial`. The current way of defining optional properties proved to be reliable during the last two years, so at least we have a solid solution.

My "I wish" is more like "I wish it would exist an official type-level operator that allows us to selectively make some key optional". The only official operator (the `?` modifier of mapped types) is applied on all keys and that's what `partial` is using under the hood.

Now back to the main topic, since there's already a way to encode optional properties, I think we are talking about API ergonomics here.

This is what you can do today

```ts
import * as t from 'io-ts'

interface Person {
  name: string
  age?: number
}

const Person = t.intersection([
  t.type({
    name: t.string
  }),
  t.partial({
    age: t.number
  })
])
```

Another option which avoids some boilerplate is to define an helper combinator

```ts
function type2<R extends t.Props, O extends t.Props>(
  required: R,
  optional: O,
  name?: string
): t.IntersectionC<[t.TypeC<R>, t.PartialC<O>]> {
  return t.intersection([t.type(required), t.partial(optional)], name)
}

// usage
const Person2 = type2(
  {
    name: t.string
  },
  { age: t.number }
)
```

Now since we are not talking about core features but.. taste? I can't judge for other people. 
Personally I'm fine with both versions. The following syntax is certainly *nice* (IMO)

```ts
const Person3 = type3({ 
  name: t.string,
  age: t.optional(t.number)
})

// versus

const Person2 = type2(
  {
    name: t.string
  },
  { age: t.number }
)
```

However I'm worried to put a possible constraint on the future development or undermine the stability of `io-ts` for a small syntax change. 

Maybe I'm too cautious but better safe than sorry, I had bad track record with conditional type in the past. /cc @sledorze

mmkal:
Thanks for the response. You are right that it would increase the possibility of a bug creeping into the implementation somewhere. So if this were to make it into the library it would have to be done thoughtfully and carefully. 

I do believe it goes a little further than taste though - I think it'd be hard to find anyone who prefers the current way of doing things from an API point of view (of course implementation is a separate matter), so taste only really comes into it in terms of how much individual people care. But there are consequences of having the "correct" way of doing things be non-intuitive/unnatural. Which is that the library tends to get used wrongly. The reason I'd love for this to make it in is that I've used io-ts on many projects, and in lots of places, people have implemented their own `optional` combinator which just does a union with `t.null` and `t.undefined`. This is 1) duplicated code and 2) can be quite dangerous, and has actually caused an outage for us in production. Here's an example of how:

```typescript
const optional = <T extends t.Type<any, any, any>>(type: T) => t.union([type, t.null, t.undefined])
const Person = t.interface({
  name: t.string,
  age: optional(t.number)
})

Person.decode({name: 'bob'}) // returns right({name: 'bob'})
Person.is({name: 'bob'}) // returns false
```

That very subtle semantic difference between `decode` and `is` may be correct but it's surprising, and a user basically has to get bitten by it to discover it. Our solution was to warn against this in documentation and try to catch it in code review. Even so, new team members would constantly accidentally try to do the same thing.

With the change to allow optional properties, the most natural and intuitive way to define partially-optional types (which in my experience, the majority of type definitions are), becomes the correct and safe way to do it too.

This is all to say, I think it's really worth the additional maintenance overhead! Please let me know if there's anything you can think of that I could help with which would make this change feel less risky. One thing that comes to mind (aside from more tests) is to mark the `optional` combinator as experimental until it's battle tested, maybe in v2 of the library, so anyone using it can assume the risk.

gcanti:
> That very subtle semantic difference between `decode` and `is` may be correct but it's surprising, and a user basically has to get bitten by it to discover it

I'd say that the difference is quite big, starting from the signatures:

while `decode` can change the result type

```ts
decode: (input: I) => Either<Errors, A>
```

`is` is just a custom type guard

```ts
is: (u: unknown) => u is A
```

Indeed you can't derive `is` from `decode` since, in general, `A` can be different from `I`/`O` 

This example shows the difference

```ts
NumberFromString.decode('1') // returns right(1)
NumberFromString.is('1') // returns false
NumberFromString.is(1) // returns true
```

which leads to

> Our solution was to warn against this in documentation and try to catch it in code review. Even so, new team members would constantly accidentally try to do the same thing.

this is a good point (and likely a documentation failure from my part), would you be willing to send a PR to improve `io-ts`'s docs, based on your experience?

Lonli-Lokli:
@gcanti maybe there is a chance/option that typescript itself can be updated to allow more simpler declaration in this library?

Issue #274, opened 2019-01-30
Strip unknown properties during encode


ggoodman:
Given a codec such as the following, I wonder if there would be a generic mechanism by which we could opt into `encode` behaviour such that undefined properties would be excluded from objects.

```js
import * as t from 'io-ts';

export const MyCodec = t.type({
  name: t.string,
});

const implementsMyCodec = {
  name: 'name',
  propertyNotDefinedInMyCodec: 'value',
};

console.log(MyCodec.encode(implementsMyCodec));
// Output includes the property propertyNotDefinedInMyCodec
```

When using `io-ts` as a data-marshalling and validation tool for inputs and outputs, I would like to make sure that I'm not accidentally leaking properties on outputs from my system. It would be great to have `io-ts` types as the source of truth for this behaviour.

ggoodman:
Off-topic: `io-ts` is a bit of a revelation to me. Keep up the great work!

gcanti:
@ggoodman I'm working on this *right now*. My plan is to deprecate the `type` combinator in favour of a `xxx` (name to be determined) combinator that strips additional properties while decoding / encoding (and making the `exact` combinator less useful). In particular I'm working on a PR that modify the internal `intersection` algorithm in order to support such a combinator.

lostintime:
@gcanti I did also hit the issue with `intersection` implementation while prototyping own `strip` combinator: looks like `intersection` is passing decoded value sequentially to next type, so intersection of 2 stripped types returns wrong result:

```typescript
const getProps = (type: t.HasProps): t.Props => {
  switch (type._tag) {
    case "RefinementType":
    case "ReadonlyType":
      return getProps(type.type)
    case "InterfaceType":
    case "StrictType":
    case "PartialType":
      return type.props
    case "IntersectionType":
      return type.types.reduce<t.Props>((props, type) => Object.assign(props, getProps(type)), {})
  }
}

export function strip<RT extends t.HasProps>(type: RT): t.Type<t.TypeOf<RT>, t.OutputOf<RT>, t.InputOf<RT>> {
  const props: t.Props = getProps(type)
  const keys = Object.keys(props)
  const len = keys.length

  return new t.Type(
    type.name,
    t.exact(type).is,
    (m, c) => type.validate(m, c).map((o: any) => {
      const r: any = {}

      for (let i = 0; i < len; i++) {
        const k = keys[i]
        if (Object.prototype.hasOwnProperty.call(o, k)) {
          r[k] = o[k]
[...]

PR #275, opened 2019-01-30
Polish: modify the implementation of intersection

gcanti:
This PR modifies the internal algorithms of `intersection` (`is`, `validate`, `decode`) in order to support a wider range of combinators.
In particular I'm interested in supporting combinators that strip additional properties while decoding / encoding.

**This is a first step towards the deprecation of both `type` and `exact`.**

Reference https://github.com/gcanti/io-ts/issues/274

/cc @ggoodman (the `strip` combinator introduced as a helper for testing purposes is a good prototype of the new API I want to add)


test/helpers.ts
```diff
@@ -82,3 +82,75 @@ export function withDefault<T extends t.Mixed>(
...
+          }
+        }
+        return errors.length ? t.failures(errors) : t.success(a)
+      }
+    },
+    a => {
+      const s: any = {}
+      for (let i = 0; i < len; i++) {
+        const k = keys[i]
+        s[k] = types[i].encode(a[k])
```
ggoodman: Could this be sped up using the `identity` check you use elsewhere?

PR #283, opened 2019-02-05
Polish: ensure brands uniqueness, fix #280

### LINKED ISSUE #280: Brands Uniqueness and Nominal typing
I just realized that new branded types (v1.8.0) are not globally unique (_nominal_), which I would probably expect from a refined type:

In the example below - 2 UserId types from different domains clashes (compiles successfully):

```typescript
import * as t from "io-ts"

export const Domain1UserId = t.brand(t.string, () => true, "UserId")
export type Domain1UserId = t.TypeOf<typeof Domain1UserId>

export const Domain2UserId = t.brand(t.string, () => true, "UserId")
export type Domain2UserId = t.TypeOf<typeof Domain2UserId>

export const rebrand = (id: Domain1UserId): Domain2UserId => id
```

Users may need to be aware of this and use a branding convention.

I'm trying to find a way to define a combinator for nominal typing - but it seem to always require explicit type:

https://github.com/gcanti/io-ts/issues/172#issuecomment-389108653

```typescript
export function nominal<A, O, I>(type: t.Type<A, O, I>): <B>() => t.Type<A & B, O, I> {
  return () => type as any
}

declare const OnesName: unique symbol

export const One = nominal(t.brand(t.string, () => true, "One"))<{ readonly [OnesName]: "One" }>()
export type One = t.TypeOf<typeof One>

// or a bit shorter ...
export const Two = nominal(t.brand(t.string, () => true, "One"))<typeof OnesName>()
export type Two = t.TypeOf<typeof Two>
```

So for every type will have to define a name symbol and This brings us back to explicit types discussion here: https://github.com/gcanti/io-ts/issues/265 :).

Using `unique symbol` in inferred positions doesn't work:

```
... error TS2527: The inferred type of 'Xxx' references an inaccessible 'unique symbol' type. A type annotation is necessary.
```

Am I missing something, is there other, more elegant way to do it?

gcanti:


gcanti:
@lostintime thanks for reviewing

PR #289, opened 2019-02-10
encode last type in union (fixes #288)

### LINKED ISSUE #288: Union does not encode last type but returns it non encoded
This is due to that line:
https://github.com/gcanti/io-ts/blob/master/src/index.ts#L1135

We discard the last codec and (i < len -1) and just return the value, without encoding it.

sledorze:
I'm not happy with the solution as the fundamental problem remains.
However it's still better than the buggy situation in which both prismatic codec at last position and stripping combinators are failing.
@gcanti ok to make a quick fix release?


gcanti:
Closing in favour of https://github.com/gcanti/io-ts/pull/290

PR #290, opened 2019-02-11
Bug Fix: align the runtime behavior of `Exact.is` with the type system

gcanti:
> only possibility would be to relax the `is` check of stripping combinators when testing for encoders

@sledorze you are right, for what concerns the `is` function, its runtime behavior **MUST** be aligned with the type system, so the proper fix of #288 is to relax its strictness. 

The same issue was already present for the deprecated `refinement`s (unions involving refinements could lead to wrong result of `encode`), while the `brand` combinator is not affected by this defect (runtime system and static system are aligned).

This change also fixes this [wrong test](https://github.com/gcanti/io-ts/blob/f1727c1edba4d03578832cd372969678fe543ea5/test/intersection.ts#L35) which should succeed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@
 **Note**: Gaps between patch versions are faulty/broken releases. **Note**: A feature tagged as Experimental is in a
 high state of flux, you're at risk of it changing without notice.
 
+# 1.8.2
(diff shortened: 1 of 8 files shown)


sledorze:
OK for me!

Issue #21, opened 2017-02-14
Overall experience with the library


gyzerok:
Hello everyone!

I'd like to wrap up here all the thoughts I have regarding the library so far which I think might worth consideration. They might be valid or not the intention is mostly to open discussion which might lead to some decisions regarding API and roadmap.

### Background

For over a year now I am playing [Elm](http://elm-lang.org/) programming language. It is quite good form different points of view, however is not applicable for every project.

For sometime now I have been developing React Native application and thus had to consider usage of TypeScript to achieve some kind of reliability close to Elm in it. One thing that was still uncovered - API data type safety. In Elm it is solved with [decoders](http://package.elm-lang.org/packages/elm-lang/core/5.1.1/Json-Decode) and I always enjoyed being able to caught errors with the data being incorrect early in my code. Thats why I was looking for something in this area and `io-ts` appeared to be brilliant solution for this problem! Than you @gcanti for making it!

### Issues with the library I have so far

Just to underline it a bit: my primary use case is typifying API responses. Especially when API is out of your control.

#### Validation vs Decoding

As I already wrote [here](https://github.com/gcanti/io-ts/pull/9#issuecomment-279629727) it seems like library already go a bit beyond just simple validation. I believe this is a good and useful feature and we can a bit change API direction to support it more seamlessly.

Let's say we do receive date as ISO string from the API. Most of the time string do not make a lot of sense for us because we want to be able work with dates. If we will remove ability to modify data during "validation" one would need to define two types within his program. One type with date being `string` which you can get immediately from your validator and one type with date being `Date` which you will return from your function.

The functionality described above is already possible, however is not really aligned with "validation" meaning and thats why might be not clear/misleading for library consumers.

One the other hand it looks like part of `io-ts` audience as @gcanti himself come to this library in order to have similar "decoding" they have in Elm/PureScript/Haskell/... so decoding will seem more familiar to them.

#### Either vs Promise

Since I am using this library for API data validation I have to use it in `Promise` context. For example:

```ts
axios
  .get(url)
  .then(res => validate(res.data, schema).fold(/* convert to Promise */)
```

Using `Either` might be more aligned to how it is done in functional languages, but in JavaScript/TypeScript land `Promise` seem to be main concept for this kind of cases. I would say that in order to simplify API and being more friendly to JS/TS people we should consider using promises as a default.

#### Make reporting nicer and part of core

[...]

PR #301, opened 2019-03-10
Add io-ts-promise to README.md community section

aeirola:
I wrote [io-ts-promise](https://github.com/aeirola/io-ts-promise) in order to make it simpler to use io-ts types at promise-based runtime boundary APIs. It was partially inspired by the discussion in #21, where the support for promises was proposed but rejected on the requirement that io-ts needs to stay synchronous.

In case the library would be useful to others as well, I'd propose to add it to the list of community packages.


gcanti:
Thanks @aeirola

PR #323, opened 2019-05-02
target fp-ts#v2-lib

leemhenson:
I know fp-ts v2 isn't baked yet, but I'm doing some monorepo investigation work and I need to get rid of fp-ts@<2 to prevent Typescript from exploding. I don't know what your plans are for io-ts & fp-ts@2 so it's up to you whether you use this branch.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "io-ts",
-  "version": "1.8.4",
+  "version": "1.8.5",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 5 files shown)


gcanti:
Thanks @leemhenson I have a local branch targeting v2 too (`fp-ts-v2`, just pushed)

> I don't know what your plans are for io-ts & fp-ts@2

The plan is: 

- first release a version which is compatible with both fp-ts@1 and fp-ts@2
- then start working on io-ts@2

leemhenson:
Is there any chance you have v2-compatible branches of:

- monocle-ts
- fp-ts-contrib
- io-ts-types

😀

gcanti:
@leemhenson for `fp-ts-contrib` I have a branch, I can work on the other two libs tomorrow

leemhenson:
That would be amazing, thanks!

PR #361, opened 2019-08-29
Either's getOrElse returns a union of the right and the result of the else

jdolle:
Minor correction to the response type of Either's `getOrElse`


jdolle:
opened against wrong repo. zz

Issue #216, opened 2018-10-16
How to describe enums and still get typescript types?


hexpunk:
I've studied the various approaches in #67 and I've done a bit of my own code to try to crack this nut, but I can't seem to figure out how to represent enums with io-ts and also get all the necessary typescript hints.  The various approaches suggested in that ticket and they work at runtime, but I lose all type hints at compile time.  The result from a `t.TypeOf` or a `decode` on my enum types only show up to typescript as `any`.  Furthermore, the values that are accepted by a function, for example, at compile time that expects the type of the `t.TypeOf` my io-ts enum accept _all_ numbers or _all_ strings instead of only the valid values associated with the given enum.

Is this a lost cause or is there some way to accurately get the full benefits of io-ts with enums?

jrunning:
Literally the first thing I tried to do after installing this module today was to try to define an enum. I'm flummoxed, and wondering if it's worth trying to move forward with io-ts at all.

omerkhawaja:
Firstly, `io-ts` library is ground-breaking.  

It's my first week using TypeScript, but how about this implementation?

```
enum TestEnum {
  Foo = 'Foo',
  Bar = 'Bar',
}

// Add additional meta data to TestEnum
const TestEnumMetaData = {
  [TestEnum.Foo]: { displayValue: 'Foo is cooooool' },
  [TestEnum.Bar]: { displayValue: 'Bar is cooooool' },
};

TestEnum.meta = TestEnumMetaData;

const TestType = t.union([
  t.literal(TestEnum.Foo),
  t.literal(TestEnum.Bar)
]);

// Type Example
const ProductV = t.type({
  id: t.number,
  name: t.string,
  quantity: t.number,
  type: TestType,
});

// Constant Example using Enum
type IProduct = t.TypeOf<typeof ProductV>;
const item: IProduct = {
  id: 3,
  name: 'yo',
  quantity: 123,
  type: TestEnum.Foo,
};

// Example to check data 
ProductV.is({
  id: 23,
  name: 'asdf',
  quantity: 1,
  type: 'Foo'
});

/**
 * Get display value from TestEnum.
 * @param {TestEnum}
 * @return {String}
 */
const getTestEnumDisplayValue = (value: TestEnum): string => {
  let result = 'None found';
  const isValid = TestType.is(value);
  if (isValid) {
    result = TestEnum.meta[value].displayValue;
  }
  return result;
}
```

noe132:
[...]

PR #366, opened 2019-09-17
Add `enum` support

alecgibson:
The `enum` is a popular aspect of TypeScript, and can be embedded in
interfaces as a concise, but descriptive, shorthand for literal string
unions:

```typescript
enum Colour {
  White = '000000',
  Black = 'ffffff'
}
```

This change adds an exported member `enum` to `io-ts`, based on
[this suggestion][1] by @noe132

It means that `enum`s can be reused directly in `io-ts`:

```typescript
const T = t.enum(Colour)
```

[1]: https://github.com/gcanti/io-ts/issues/216#issuecomment-471497998


alecgibson:
One place this does seem to fall down is that you lose the casting when extracting an interface. I can't quite figure out how to do that (possibly something to do with the `encode` function?). For example:

```typescript
enum Colour {
  White = '000000'
}

const Style = t.type({
  colour: t.enum(Colour)
})

interface IStyle extends t.TypeOf<typeof Style> {}

const style: IStyle = {
  colour: Colour.White
}

const colour = style.colour // colour has type "any" rather than "Colour"
```

I tried tweaking the signature for `enumType`:

```typescript
const enumType = <E, A = E[keyof E]>(e: E, name: string = 'Enum'): EnumType<A> => {
  const is = (u: unknown): u is A => Object.keys(e).some((k) => e[k as keyof E] === u)
  return new EnumType<A>(name, is, (u, c) => (is(u) ? success(u) : failure(u, c)), identity)
}
```

...which seems to only succeed in disallowing a `decode` to `string` (which I guess is correct?), but still doesn't correctly cast.

jessmorecroft:
This looks like it might work better, for string enums at least...

```
const enumType = <E>(
  e: { [key: string]: E },
  name: string = 'Enum',
): EnumType<E> => {
  const is = (u: unknown): u is E => Object.keys(e).some(k => e[k] === u);
  return new EnumType<E>(
    name,
    is,
    (u, c) => (is(u) ? t.success(u) : t.failure(u, c)),
    t.identity,
  );
};
```

saa18877433:
Re: [gcanti/io-ts] Add `enum` support (#366)

2020-02-16 2:56 غرينتش-08:00, jessmorecroft <notifications@github.com>:

noe132:
Hey, since number enums have remapping,
https://www.typescriptlang.org/docs/handbook/enums.html#reverse-mappings
I think it's essential to check if the key is the remapped one.
```typescript
enum E {
  A = 1
}

// compiles to
const E {
  A: 1,
  '1': 'A',
}
```
In this case, `1` is valid, but `'A'` is not.
https://www.typescriptlang.org/play?#code/FAUwHgDg9gTgLgAhAOwK4FsEFEEG9gIICCCAvAgIzAC+QA

I've updated my code here https://github.com/gcanti/io-ts/issues/216#issuecomment-471497998

abrkn:
Any update on this?

aroman:
+1 this would be a great addition, as currently I need to manually copy/paste my enum values into a `t.union` of `t.literal`s

alecgibson:
Okay I've updated this, and it now correctly casts!

Screen Shot 2021-01-04 at 16 40 49

Also updated to include @noe132 's update for checking for reverse mapping

Andrapyre:
Any update on this from @gcanti ?

halbich:
Hi, what is the current status of this feature? We'd love to have enum support :)

luisroel91:
@gcanti How do you feel about merging this in?

martindines:
Is there anything stopping this feature from being merged?

PR #367, opened 2019-09-17
Add `Record` shorthand to `README`

alecgibson:
When writing an `interface`, consumers of `io-ts` may more readily
recognise the `Record` shorthand syntax, without realising that it's
actually a `Record` (which is an [advanced type][1]):

```typescript
interface Example {
  [key: string]: string;
}
```

This change makes a small update to the "record of type" entry in the
`README` table, so that this feature may be more discoverable.

[1]: https://www.typescriptlang.org/docs/handbook/advanced-types.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,34 +289,34 @@ You can also use the [`withMessage`](https://gcanti.github.io/io-ts-types/module
 
 # Implemented types / combinators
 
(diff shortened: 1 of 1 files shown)


alecgibson:
Closing, since the `README` has now radically changed.

PR #372, opened 2019-10-17
Fix tuple decode, fail on additional components

DenisFrezzato:
This PR fix the `decode` method of `tuple`, failing with additional components (as the `is` method does and as the type system behaves).


DenisFrezzato:
Closing this, it looks like is not a welcome change :)

PR #427, opened 2020-02-29
Rewrite some type definition with strict index access

nicu-chiciuc:
Great package, I've been testing it on some internal lambda functions and it seems to make interfaces much safer.
I'm not sure if the current commit will align with the direction of the package but I would really appreciate if you would take it into consideration.

So the problem, as you may know, is that:
```ts
const arr: number[] = [];
const num: number = arr[0];
```
Is completely valid and TS will not show any errors even though it is clearly incorrect.
This problem is discussed here: https://github.com/microsoft/TypeScript/issues/13778

In our project, we're starting to try out this approach by using `patch-package` to minimize any possible run-time errors.
Basically changing:
```ts
interface Array {
// ...
  [n: number]: T;
}
```
to
```ts
interface Array {
// ...
  [n: number]: T | undefined;
}
```

Unfortunately I noticed that our branch on which `io-ts` is used there are some type errors created by this change. TS cannot infer that checking that given `arr.length > 1` implies `arr: [T, T, ...T[]]`. 

So things like:
```ts
export interface IntersectionC<
// ...
CS extends { length: 2 }
      ? TypeOf<CS[0]> & TypeOf<CS[1]>
// ...
```
Could not be proven to TS as correct.
All this commit does is change some problematic definition to the following form:
```ts
export interface IntersectionC<
// ...
CS extends [Mixed, Mixed]
      ? TypeOf<CS[0]> & TypeOf<CS[1]>
// ...
```

At times this requires writing more code since TS doesn't have higher-order types but I think this is a small change that would be greatly appreciated by us.

### DIFF
--- a/docs/modules/index.ts.md
+++ b/docs/modules/index.ts.md
@@ -385,13 +385,7 @@ Added in v1.8.1
 **Signature**
 
 ```ts
-export interface IntersectionC<
-  CS extends
-    | [Mixed, Mixed, ...Mixed[]]
-    | [Mixed, Mixed, Mixed, ...Mixed[]]
-    | [Mixed, Mixed, Mixed, Mixed, ...Mixed[]]
-    | [Mixed, Mixed, Mixed, Mixed, Mixed, ...Mixed[]]
->
+export interface IntersectionC<CS extends [Mixed, Mixed, ...Mixed[]]>
   extends IntersectionType<
     CS,
     CS extends [Mixed, Mixed]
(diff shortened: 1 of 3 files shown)


saa18877433:
في السبت، ٢٩ فبراير، ٢٠٢٠ ٤:١١ م Chiciuc Nicușor <notifications@github.com>
كتب:

gcanti:
> In our project, we're starting to try out this approach by using patch-package to minimize any possible run-time errors.

@nicu-chiciuc I'm sorry, I don't want to argue about the correctness of this approach, you are free to experiment of course, but I'm not going to change the typings for such a use case

nicu-chiciuc:
I understand. Just wanted to mention that even if one doesn't change the indexing definition, the typings will work exactly as they did before.

PR #435, opened 2020-04-01
Handle missing properties correctly

### LINKED ISSUE #434: Handle missing properties correctly
### Current Behavior

```typescript
const T = t.type({ a: t.unknown })
assert.strictEqual(T.is({}), true) // property 'a' is missing, but the type check still passes
```

### Expected behavior

```typescript
const T = t.type({ a: t.unknown })

// property 'a' is missing but must be present
assert.strictEqual(T.is({}), false) 

// property 'a' is present, and undefined is a valid value for unknown
assert.strictEqual(T.is({ a: undefined }), true) 

// property 'a' is present, and null is a valid value for unknown
assert.strictEqual(T.is({ a: null }), true) 
```

Similarly, in vanilla TypeScript:
``` typescript
type T = { a: unknown }

// won't compile
const T1: T = { } 

// good
const T2: T = { a: undefined } 

// good
const T3: T = { a: null } 
```

### Reproducible example

See above

### Suggested solution(s)

Will submit a pull request.

### Your environment

| Software   | Version(s) |
| ---------- | ---------- |
| io-ts      |   master   |
| fp-ts      |   2.0.0         |
| TypeScript |   3.7.4     |

leilapearson:
Resolves #434 

Changes the handling of missing properties in `t.type` interface types to align with TypeScript.

Adds many additional tests (~170) to confirm that everything behaves as expected.  Also changes a small number of pre-existing tests that asserted the old behavior.

Note that people who rely on the old behavior may see this fix as a breaking change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "io-ts",
-  "version": "2.0.6",
+  "version": "2.1.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -2217,7 +2217,8 @@
           "version": "2.1.1",
           "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
           "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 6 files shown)


SHaTRO (commented):
Declaring something as "undefined" is semantically equivalent to declaring it as missing.
One should use "null" to declare a property as "exists (can exist) but with no value".

This PR does not appear to respect those semantics and therefore is definitively a breaking change and inconsistent with the semantic definition of "undefined".

PR #460, opened 2020-05-02
add `Eq` experimental module

gcanti:


test/Eq.ts
```diff
@@ -0,0 +1,81 @@
...
+    const eq = E.UnknownArray
+    assert.deepStrictEqual(eq.equals(['a'], ['a']), true)
+    assert.deepStrictEqual(eq.equals(['a'], ['b']), true)
+    assert.deepStrictEqual(eq.equals(['a'], ['a', 'b']), false)
+  })
+
+  it('UnknownRecord', () => {
+    const eq = E.UnknownRecord
+    assert.deepStrictEqual(eq.equals({}, {}), true)
+    assert.deepStrictEqual(eq.equals({ a: 1 }, { a: 1 }), true)
```
mmkal, line 23: From the implementation, it looks like only keys are looked at, not values. If that's right could there be a test here to make that explicit/deliberate?
```suggestion
    assert.deepStrictEqual(eq.equals({ a: 1 }, { a: 1 }), true)
    assert.deepStrictEqual(eq.equals({ a: 1 }, { a: 2 }), true)
```

On that note, it might be good to add something to the docs re why it works that way, since it looks counterintuitive/different from a generic `isEqual` function for arbitrary dictionaries e.g. lodash.isEqual.

aabccd021:
Will this be merged with the index.ts module?
I would love it if in future we can do:
```ts
import * as t from 'io-ts'

const Person = t.type({
  name: E.string,
  age: E.number
})

console.log(Person.equals({ name: 'a', age: 0 }, { name: 'a', age: 0 })) // => true
```

PR #474, opened 2020-05-27
Use data-last chain & map in index

mikearnaldi:
This PR removes the dependency on the `either` instance as a mean of improving tree shaking


gcanti:
@mikearnaldi as discussed offline https://github.com/gcanti/io-ts/pull/475

PR #507, opened 2020-08-08
Bare imports without lib/ and es6/ directories

### LINKED ISSUE #506: Import without /lib or /es6 paths like fp-ts now does
### Current Behavior

To import specific files in `io-ts`, you need to import with `/lib/` or `/es6/` prefixes, like `import { Encoder } from "io-ts/lib/Encoder"`

### Desired Behavior

Don't require that prefix, like `import { Encoder } from "io-ts/Encoder"`

### Suggested Solution

Do what `fp-ts` is now doing with https://github.com/gcanti/fp-ts/pull/1241 to simplify imports

### Who does this impact? Who is this for?

All users who use sub-package imports

### Describe alternatives you've considered

handling doing the correct import manually leads to a lot of extra boilerplate in setting up linters to avoid the wrong imports, getting confusing errors when you use the wrong one.

### Additional context

`fp-ts` already shows a proof of concept so would be nice to apply it here.

### Your environment

| Software   | Version(s) |
| ---------- | ---------- |
| io-ts      | 2.2.9           |
| fp-ts      | 2.8.1           |
| TypeScript | 3.9.6           |

osdiab:
Similar to https://github.com/gcanti/fp-ts/pull/1241/files and https://github.com/gcanti/io-ts-types/pull/137 , adds support for importing from `io-ts/Decoder`, etc. instead of `/lib/Decoder` and so on. Just copied logic over from `fp-ts`. Thanks for the great projects @gcanti !

Closes #506


package.json
```diff
@@ -38,7 +37,7 @@
   "homepage": "https://github.com/gcanti/io-ts",
   "dependencies": {},
   "peerDependencies": {
-    "fp-ts": "^2.0.0"
+    "fp-ts": "^2.8.0"
```
mmkal, line 40: Does this qualify as a breaking change?

gcanti: @mmkal yes it does

@osdiab if I'm not mistaken we can avoid such a breaking change by keeping the `import-path-rewrite` step

osdiab: Cool, redid it without removing `import-path-rewrite`! Thanks for the tip.

osdiab:
@gcanti I made it not backwards incompatible - any next steps necessary?

gcanti:
@osdiab LGTM, thanks!

gcanti (approved):

PR #509, opened 2020-08-15
docs: fix Codec.md

SeungUkLee:
An `"error TS2322: Type 'Codec<unknown, string, unknown>' is not assignable to type 'Codec<unknown, string, number>'.
  Type 'unknown' is not assignable to type 'number'"` error occurs when using the example source code of the [`Codec.md`](https://github.com/gcanti/io-ts/blob/master/Codec.md) file.

So I fix `encoder` type from `E.Encoder<string, unknown>` to `E.Encoder<string, number>`

### DIFF
--- a/Codec.md
+++ b/Codec.md
@@ -38,7 +38,7 @@ const decoder: D.Decoder<unknown, number> = pipe(
   })
 )
 
-const encoder: E.Encoder<string, unknown> = {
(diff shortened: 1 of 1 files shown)


SeungUkLee:
Sorry. I used example source code(in Codec.md) incorrectly. So I close this PR. :sob:

Issue #483, opened 2020-06-25
Support for excluding undefined through intersection


cyberixae:
When [generating codecs from JSON schemas](https://www.npmjs.com/package/io-ts-from-json-schema) it seems useful to define field types and required fields separately.
To do this I would need a codec for `Exclude<unknown, undefined>`.

```typescript
type Example = Partial<{
  foo: 123
  bar: 456
}>
type RequireFoo = Example & {
  foo: Exclude<unknown, undefined>
}
const requireFoo: RequireFoo = {
  foo: 123
}
```

cyberixae:
Actually `Exclude<unknown, undefined>` doesn't seem to work but `{}|null` would perhaps also do the job.

```
type Example = Partial<{
  foo: 123
  bar: 456
}>
type Defined = {}|null
type RequireFoo = Example & {
  foo: Defined
}
const requireFoo: RequireFoo = {
  foo: 123
}
```

cyberixae:
I'm trying `{}|null` at maasglobal/io-ts-from-json-schema#19

PR #513, opened 2020-08-28
add defined value codec

cyberixae:
This PR adds a codec for defined values, the opposite of `undefined`. It is useful for situations where you have to describe the exact type and assert the existence of the type separately. See #483


cyberixae:
After playing around with this for some time, I think it doesn't work well enough to be part of the main package. Perhaps it could be added into contrib. I'll consider reopening the PR there.

PR #520, opened 2020-10-03
Make toForest stack-safe

safareli:
Without this change on big input with many errors toForest throws stack overflow.

Screen Shot 2020-10-03 at 8 40 50 PM

---

I've added very simple test for FreeSemigroup because of this:
Screen Shot 2020-10-03 at 8 48 44 PM


gcanti:
Thanks @safareli, released.

safareli:
🎉

PR #522, opened 2020-10-11
chore (docs): removed pipe from intersect in docs

enricopolanski:
Thought of removing `pipe` as I found its use unnecessary for this binary combinator.


enricopolanski:
Closing it as `pipe` does offer a better type inference.

PR #540, opened 2020-12-13
Experimental: add `getErrorForest`

jituanlin:
Add `getErrorForest` for customize error report.
Errors tree can be got by `getErrorForest` to customize error report

```ts
 const decoder = _.type({
    a: _.string
  })
 const tree = getErrorForest(decoder.decode({ c: [1] })
/*
E.left([
  {
    value: {
      _tag: 'Key',
      key: 'a',
      kind: 'required'
    },
    forest: [
      {
        value: {
          _tag: 'Leaf',
          error: 'string',
           actual: undefined
        },
        forest: []
      }
    ]
  }
])
*/
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules
 dev
 coverage
 declaration/out/src
+/.idea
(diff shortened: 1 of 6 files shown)


jituanlin:
This implementation is not stack safe, close it... ...

PR #555, opened 2021-01-21
Remove non null assertion in Decoder.toForest

safareli:


gcanti:
Awesome, thank you @safareli

PR #557, opened 2021-01-27
doc: exclude zero from set of positive integers

EricCrosson:


EricCrosson:
This actually introduced a bug into our codebase, a new employee's first codec was a copy/paste of this one :sweat_smile:

PR #562, opened 2021-02-03
Fix.unknown record

### LINKED ISSUE #559: Record<string,unknow> fails on non object literals
### Current Behavior

Returns `Left` when using the `decode` function.

### Expected behavior

Returns `Right` when using the `decode` function.

### Reproducible example
```ts
const element = document.body

const result = c.UnknownRecord.decode(element)
// returns: `Left`
```

### Suggested solution(s)

1.
```diff
 exports.UnknownRecord = {
-    is: function (u) { return Object.prototype.toString.call(u) === '[object Object]'; }
+    is: function (u) { return Object.prototype.toString.call(u).startsWith('[object '); }
 };
```

### Additional context

I'm not using `io-ts` to deal with HTTP responses, I'm using it to validate some `Event`'s that have been created with the custom event API.

```
const event = new Event()

const result = Object.prototype.toString.call(event)
// returns: "[Object Event]"
```

### Your environment

Which versions of io-ts are affected by this issue? Did this work in previous versions of io-ts?

| Software   | Version(s) |
| ---------- | ---------- |
| io-ts      |            |
| fp-ts      |            |
| TypeScript |            |

waynevanson:
I've added some test against the `Stream` object in node, to demonstrate that it'll accept any non-primitive, non-null, non-array object.

closes #559

### DIFF
--- a/src/Guard.ts
+++ b/src/Guard.ts
@@ -93,7 +93,7 @@ export const UnknownArray: Guard<unknown, Array<unknown>> = {
  * @since 2.2.0
  */
 export const UnknownRecord: Guard<unknown, Record<string, unknown>> = {
-  is: (u: unknown): u is Record<string, unknown> => Object.prototype.toString.call(u) === '[object Object]'
+  is: (u: unknown): u is Record<string, unknown> => u != null && typeof u === 'object' && !Array.isArray(u)
 }
 
 // -------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


waynevanson:
I saw you pushed a commit, but check out my comment on your commit: https://github.com/gcanti/io-ts/commit/d45d063f705229eedb0bea4d8e7bbf23e31eea98#r46684107

PR #567, opened 2021-02-17
version 2.2.15

gcanti:


fbunau:
I'm pretty new to this library, so excuse the possible ignorance.

Can this new readonly combinator be used to add "_tag" values ? 
I really like for the stable part of the library that you can use "io-ts-extra" for pattern matching. But I am completely lost on how this would work with the new Decoder. I tried adding _tag, but can't get it to work, as I want it to just be a tag, not a decoding requirement. I'm thinking this new combinator would help with that ?

gcanti:
> Can this new readonly combinator be used to add "_tag" values ?

No, the `readonly` combinator just makes the `A` type readonly:

```ts
import * as D from 'io-ts/Decoder'

const d1 = D.struct({
  a: D.string
})
type D1 = D.TypeOf<typeof d1>
/*
type D1 = {
    a: string;
}
*/
const d2 = D.readonly(d1)
type D2 = D.TypeOf<typeof d2>
/*
type D2 = {
    readonly a: string;
}
*/
```

> I am completely lost on how this would work with the new Decoder

Ongoing discussion [here](https://github.com/gcanti/io-ts/issues/453#issuecomment-755318009)

fbunau:
Thanks for the quick reply. I just realized that Decoder has the "sum" (tagged union) combinator. It does what I was looking for.

// readonly combinator is good. I use these models for data classes, all fields are used in a readonly way. good usecase :+1:

PR #584, opened 2021-04-17
POC: prune warnings

0x706b:
An experiment on pruning warnings from intersections in the poc. As a note, there are some pieces missing and some cases not covered, and it's not stack-safe, but it appears to work with some simple examples. This method also only works with string keys so far, but I imagine could be slightly modified to work with numeric keys as well.


0x706b:
closing in favor of current solution

Issue #485, opened 2020-06-25
How to model objects regardless of their constructor?


OliverJAsh:
### Current Behavior

I am trying to create a runtime type for `{ statusCode: number }`, however if I use `t.type`/`t.interface` it won't work if I try to use it to decode an object that has type `Error & { statusCode: number }`:

```ts
import * as t from 'io-ts';

const StatusCodeObject = t.type({ statusCode: t.number });

{
    const error = { statusCode: 200 };
    console.log(StatusCodeObject.is(error)); // => true
}

{
    const error = new Error('foo') as Error & { statusCode: number };
    error.statusCode = 200;
    console.log(StatusCodeObject.is(error)); // => false
}
```

Is there any way to write `StatusCodeObject` so that this will work? Essentially I want the runtime type to work with objects of any type, including things like `Error`, `Date`, or anything else that could have the property I'm looking for.

Reading https://github.com/gcanti/io-ts/issues/413, I understand that io-ts is designed so that `t.type` should not work with `Error`. In that case, perhaps we need another way to model this?

### Your environment

Which versions of io-ts are affected by this issue? Did this work in previous versions of io-ts?

| Software   | Version(s) |
| ---------- | ---------- |
| io-ts      |            |
| fp-ts      |            |
| TypeScript |            |

VanTanev:
@OliverJAsh As a stopgap, you could write a `typeAny()` variation of `t.type()` that is something like:

```typescript
const toSimpleObj = (o: any): object => JSON.parse(JSON.stringify(o))

export const typeAny = <P extends t.Props>(
    props: P,
    name = `${getNameFromProps(props)}`,
): t.TypeC<P> => {
    const type = t.type(props)

    return new t.InterfaceType(
        name,
        (u): u is {[K in keyof P]: t.TypeOf<P[K]>} => type.is(toSimpleObj(u)),
        (u, c) => type.validate(toSimpleObj(u), c),
        a => type.encode(a),
        props,
    )
}

const getNameFromProps = (props: t.Props): string =>
    Object.keys(props)
        .map(k => `${k}: ${props[k].name}`)
        .join(', ')
```

PR #611, opened 2021-10-12
Allow Error to be decoded with UnknownRecord 

louix:
Some libraries (e.g. AWS SDK, Axios) will throw Errors, but won't export the actual Error itself which prevents an  `e instanceof XError` check.

It's useful to be able to write a codec for these Errors, but no longer works after `2.0.5` ([previous implementation](https://github.com/gcanti/io-ts/blob/18e6437bbb9bc63c3ecfc6a1361eaa6250a9ce38/src/index.ts#L421)).

Related to issue #485.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -993,7 +993,7 @@ export class AnyDictionaryType extends Type<{ [key: string]: unknown }> {
       'UnknownRecord',
       (u): u is { [key: string]: unknown } => {
         const s = Object.prototype.toString.call(u)
-        return s === '[object Object]' || s === '[object Window]'
+        return s === '[object Object]' || s === '[object Window]' || s === '[object Error]'
       },
       (u, c) => (this.is(u) ? success(u) : failure(u, c)),
       identity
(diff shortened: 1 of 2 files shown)


bradleyayers:
I'm running into this issue now, it was very surprising to find that `UnknownRecord` doesn't consider instances of `Error` to be compatible.

Concretely, the following returns false:

```js
t.type({}).is(new Error()); // false
```

Is there a compelling reason for why Error should not be included here?

GriffinSchneider:
I also ran into this issue, we switched to a fork with this PR applied which is working well so far.

The fix in this PR still allows some surprising behavior on weird objects though, since `Object.prototype.toString` can be customized like this:

```typescript
const obj = { [Symbol.toStringTag]: 'something' };
Object.prototype.toString.call(obj); // '[object something]'
```

It looks like the original intention of [the change in 2.0.5 ](https://github.com/gcanti/io-ts/commit/231b9f58218fc5e3f64ce28a183c81afff1fb642) was just to exclude arrays from `UnknownRecord`. The most reliable way to do that is probably to exclude arrays explicitly, rather than trying to build a list of non-array types that are allowed. For example, with `Array.isArray`:

```typescript
u !== null && typeof u === 'object' && !Array.isArray(u)
```

Issue #542, opened 2020-12-14
Make optional fields more user friendly


safareli:
Other decoding/validation libraries have more user friendly optionality support like this:
```ts
D.type({
  foo: D.string,
  baz: D.optional(D.string)
})
```

Would be great to have something like this here too. As discussed with @gcanti there was some issues with the stable API #140 #266 but he doesn't remember if he even tried this with the experimental API.

aroman:
Any update on this? Would be very helpful.

ruizb:
The only "elegant" way I found recently is to use `optionFromNullable` from the [io-ts-types](https://github.com/gcanti/io-ts-types) library. However, it uses an `Option<A>` instead of `A | undefined`, so maybe it's not exactly what you are looking for...

```ts
import * as t from 'io-ts'
import { optionFromNullable } from 'io-ts-types'

const Foo = t.type({
  foo: t.string,
  baz: optionFromNullable(t.string)
})

type Foo = t.TypeOf<typeof Foo> // { foo: string, baz: O.Option<string> }

console.log(Foo.decode({ foo: 'foo' }))             // right({ foo: 'foo', baz: none })
console.log(Foo.decode({ foo: 'foo', baz: 'baz' })) // right({ foo: 'foo', baz: some('baz') })
```

| Software | Version |
| --- | --- |
| fp-ts | 2.9.5 |
| io-ts | 2.2.16 |
| io-ts-types | 0.5.15 |

safareli:
After being inspired by [sparceType from io-ts-extra](https://github.com/mmkal/ts/blob/c7a416c8f04f41f6e977f2af0c37001132b45f0f/packages/io-ts-extra/src/combinators.ts#L7-L68) I've implemented similar thing for Decoder:

```ts
import * as D from "io-ts/Decoder";
import { pipe } from "fp-ts/lib/function";
import { partition } from "fp-ts/lib/Record";

type AnyDecoder = D.Decoder<unknown, unknown>;

type Props = { [K in string]: AnyDecoder & Partial<Optional> };

export interface Optional {
  optional: true;
}

const isOptional = <T>(val: T & Partial<Optional>): val is T & Optional => {
  return val.optional ?? false;
};

type OptionalKeys<Base> = {
  [Key in keyof Base]: Base[Key] extends Optional ? Key : never;
}[keyof Base];

type RequiredKeys<Base> = {
  [Key in keyof Base]: Base[Key] extends Optional ? never : Key;
}[keyof Base];

[...]

PR #628, opened 2022-01-28
Add optional

thewilkybarkid:
This adds `optional` to the `Decoder`, `Encoder` and `Codec` modules, which handles conversion between `undefined` and `Option`.

This is a little different to [`optionFromNullable`](https://github.com/gcanti/io-ts-types/blob/master/src/optionFromNullable.ts) in that it _only_ works with `undefined` and not `null` (`optionFromNullable` decodes from both, and encodes to `null`).

Refs https://github.com/gcanti/io-ts/issues/542#issuecomment-1010844123.

### DIFF
--- a/docs/modules/Codec.ts.md
+++ b/docs/modules/Codec.ts.md
@@ -34,6 +34,7 @@ Added in v2.2.3
   - [lazy](#lazy)
   - [mapLeftWithInput](#mapleftwithinput)
   - [nullable](#nullable)
+  - [optional](#optional)
   - [partial](#partial)
   - [readonly](#readonly)
   - [record](#record)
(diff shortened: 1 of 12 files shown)


thewilkybarkid:
Just tried this with the `json` Encoder from #638 and found that the `undefined` is not assignable to `Json` (i.e. it's not quite the same as not existing in the type).

thewilkybarkid:
Options I can see are to make the `JsonRecord` type have `undefined` properties, or to use something like https://github.com/Microsoft/TypeScript/issues/12400#issuecomment-758523767 on the `struct` types (i.e. to make properties that can have `undefined` as a value optional).

manocha-aman (approved):

PR #642, opened 2022-04-17
Explicit Error Types

anthonyjoeseph:
Completes the work started by @gcanti in the [poc branch](https://github.com/gcanti/io-ts/tree/poc), which is explained at a high level [here (by the author)](https://github.com/gcanti/io-ts/issues/453#issuecomment-842095270) and [in the docs](https://github.com/anthonyjoeseph/io-ts/blob/poc-complete/Decoder.md)

As per the notes in the [POC branch](https://github.com/gcanti/io-ts/blob/poc/src/Schemable2.ts#L15), all `Schemable`-related code has been moved into a separate repo called [io-ts-contrib](https://github.com/anthonyjoeseph/io-ts-contrib). It's registered under [anthonyjoeseph's GitHub account](https://github.com/anthonyjoeseph), but ownership would ideally be transferred to a maintainer of io-ts

(This PR is still a draft, pending the following)

# TODO
- Update Decoder readme to explain how to add custom `LeafE` error types that are printable (based on [gcanti's explanation](https://github.com/gcanti/io-ts/issues/453#issuecomment-842095270))
- Complete test coverage (currently only [old v2.2 tests](https://github.com/gcanti/io-ts/blob/master/test/Codec.ts) have been ported up to TaskDecoder & Codec, new [Decoder2 function tests](https://github.com/gcanti/io-ts/blob/poc/test/Decoder2.ts) have yet to be ported)
- Figure out how to use Codec.lazy w/o @ts-expect-error
- Replace snapshot testing (compare printed string errors) (?)
- Reduce fp-ts version (?)

# io-ts-contrib TODO:
- Docs
  - `Guard`
  - `Arbitrary`
  - `JsonSchema`
  - `Type`
- Complete test coverage

# Open Questions

Does io-ts 3.0.0 depend on fp-ts 3.0.0 still? (I thought I saw that somewhere but I can't find it now) Why?

Any way to use [Kleisli](https://github.com/gcanti/io-ts/blob/master/src/Kleisli.ts) to remove duplicate code between Decoders and TaskDecoders? Maybe through the help of multiple case-specific Applicative<These>, Chain<These>, Semigroup<These> and/or Alt<These> instances? (Or is this even worth doing?)

Why do decoders store their metadata at the value level? DecodeErrors make sense, that metadata is used for walking the tree, but what's the decoder stuff for?

Why was 'refine' originally removed from Decoder2? fromRefinement? (This PR adds `refinement` back in)

Add Alt & Category instances for Decoder and TaskDecoder?

Decoder/TaskDecoder - input Ord instance for traversing? Or standardize "Ord.trivial"?
  - this would mean we'd have to bump fp-ts version (?)

TaskDecoder - input Applicative instances? Or standardize "Task.parallel" (would probably be sufficient for e.g. graphql resolvers)?


src/Codec.ts
```diff
@@ -8,364 +8,425 @@
...
- */
-export function fromDecoder<I, A>(decoder: D.Decoder<I, A>): Codec<I, A, A> {
-  return {
-    decode: decoder.decode,
-    encode: identity
-  }
+export interface Codec<D, E> {
+  readonly decoder: D
+  readonly encoder: E
 }
```
thewilkybarkid, line 26: Based on

> A `Codec` is just a `Decoder` and its dual packed together

this isn't strictly typed.

It could be

```ts
export interface Codec<D extends D.Decoder<unknown, unknown, unknown>, E extends D.Decoder<unknown, unknown, unknown>> {
  readonly decoder: D
  readonly encoder: E
}
```

but it still means the type of the decoder and input to the encoder is different. So, maybe:

```ts
export interface Codec<D extends D.Decoder<unknown, unknown, unknown>, E extends D.Decoder<D.TypeOf<D>, unknown, unknown>> {
  readonly decoder: D
  readonly encoder: E
}
```

Some existing usage starts to fail with these; I've not tried to see why yet.

thewilkybarkid: > These are ultimately solvable problems that would require shifting the implementation of Codec to use its own idioms, separate from the implementation of `Decoder`. Alternately, the `Decoder` module could be re-written to accommodate this

I do wonder if the API should retain `encode`/`decode` rather than `encoder`/`decoder` (`C.string.encoder.decode` looks a little confusing). It does change what it is though.

> My guess is that a `Codec` is meant to simply provide combinators for modifying two Decoders in tandem, giving the user a bit of flexibility as to how those decoders actually relate to each other. I added that line to the documentation to explain why someone might _want_ to use a Decoder, but this is a good point - that specific phrasing is technically incorrect. Maybe something more like 'a Codec is two decoders packed together, the second often undo-ing the change made by the first'?

That could work ('reversing'?). Maybe it is fine that it isn't typed: feels a bit odd to have one that doesn't work on the same type, but if there isn't a technical reason to enforce it...

anthonyjoeseph: Oh I understand - I think it might help for them to conform to the `Decoder` interface for the sake of the existing [Decoder type functions](https://github.com/gcanti/io-ts/pull/642#discussion_r853066866)

In any case, would you mind if I mark this conversation as resolved? I'm not super comfortable making changes here that change the intent of the [original poc branch](https://github.com/gcanti/io-ts/tree/poc), not at least without input from the author

cdimitroulas:
hey @anthonyjoeseph, it's been a while since this was opened. Is it still the direction of travel for the io-ts POC and (hopefully) next major version of it? Do you need any help with this?

PR #649, opened 2022-07-16
Add usage to Readme

koenigstag:
Added simple usage block
```ts
import * as t from 'io-ts'

const User = t.type({
  userId: t.number,
  name: t.string
})
```


EricCrosson:
Is this not already covered by the "Example" section of this page? https://github.com/gcanti/io-ts/blob/master/index.md

PR #653, opened 2022-08-02
Fix UnkownRecord.is on undefined

### LINKED ISSUE #652: UnknownRecord.is allows undefined, which leads to exception in nonEnumerableRecord function
## 🐛 Bug report

UnknownRecord.is allows undefined which leads to an exception inside the library. The exception occurs in nonEnumerableRecord function when calling Object.keys on undefined that passed UnkownRecord.is.
https://github.com/gcanti/io-ts/blob/920b21ea32e82b328cd14ff4b28da72ae60db505/src/index.ts#L406-L409

### Current Behavior

`UnknownRecord.is(undefined)` is `true`. Because of the false type guard type errors can occur. See the real-world example above.

### Expected behavior

`UnknownRecord.is(undefined)` should be `false`. 

### Suggested solution(s)

UnknownRecord.is should also check for undefined as it checks for null here:
https://github.com/gcanti/io-ts/blob/920b21ea32e82b328cd14ff4b28da72ae60db505/src/Guard.ts#L96

vendi95:
Fixes #652

### DIFF
--- a/src/Guard.ts
+++ b/src/Guard.ts
@@ -93,7 +93,7 @@ export const UnknownArray: Guard<unknown, Array<unknown>> = {
  * @since 2.2.0
  */
 export const UnknownRecord: Guard<unknown, Record<string, unknown>> = {
-  is: (u: unknown): u is Record<string, unknown> => u !== null && typeof u === 'object' && !Array.isArray(u)
+  is: (u: unknown): u is Record<string, unknown> => u !== null && u !== undefined && typeof u === 'object' && !Array.isArray(u)
 }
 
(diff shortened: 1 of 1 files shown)


Avaq:
That's odd. Doesn't `typeof undefined` return `'undefined'`? Shouldn't that mean that the added condition in this PR is redundant with the one after it?

vendi95:
> That's odd. Doesn't `typeof undefined` return `'undefined'`? Shouldn't that mean that the added condition in this PR is redundant with the one after it?

You are right. This can not be the problem.

Issue #450, opened 2020-04-19
A partly partial with io-ts


james2mid:
I just posted a new question [on StackOverflow](https://stackoverflow.com/questions/61311182/a-partly-partial-with-io-ts) regarding making a partial from an existing schema.

Any help would be appreciated.

---

I'm having some issues with using `io-ts`. I've found that its really lacking documentation, most of the progress I've made is through GitHub issues. And no I don't understand the HKTs so no help there.

Basically I'm creating a type elsewhere, `type` which has a certain set of keys:
```ts
import * as S from 'io-ts/lib/Schema'

const someType = S.make(S => S.type({
  id: S.string,
  foo: S.string,
  bar: S.boolean
}))

type T = S.TypeOf<typeof someType>
```

And I need to transform this to a *partial* where some keys are still required. Say just the `id` key is still required, defining the type in TS would be like:
```ts
type PartlyPartial <T> = Partial<T> & { id: string }
```
Simple!

Now, I want to be able to do this at runtime through `io-ts` so that I can create a validator function with the `Guard` module. This is what I've got so far:
```ts
import * as G from 'io-ts/lib/Guard'

const propsBase = S.make(S => S.type({
  id: S.string
}))

const partial = S.make(S => S.intersection(
  someType(S) // needs to be partial,
  propsBase(S)
))

// So that this returns `true`
partial(G.guard).is({
  id: 'some id',
  bar: true
})
```

But the `someType(S)` defines the type as having all keys required, but none of them should be required.

gcanti:
> making a partial from an existing schema

That's not possible

james2mid:
I appreciate you getting back to me.

Is there no workaround that I could create temporarily? All I need to do it validate with these conditions.

Could the `properties` be attached as a property to the Schema fn produced by `make` for future manipulation. I understand it seems you've tried to get away from the whole `props` idea after dealing with io-ts' initial `Type` but in my case it seems necessary.

RemiLelaidier:
Found that
https://stackoverflow.com/questions/61311182/a-partly-partial-with-io-ts

james2mid:
> Found that
> https://stackoverflow.com/questions/61311182/a-partly-partial-with-io-ts

Yes, that's my post which I linked at the beginning, although I didn't reply saying that I found a workaround.

If anybody has the same issue, then its possible to create a wrapper around `make` which provides the features I'm talking about.

PR #654, opened 2022-08-19
adding SemiPartialType and associated tests

mjburghoffer:
Since having a struct with a mix of required and optional properties is so common, I figured it would be nice (and more performant) to have a purpose-built type.  I know you could achieve similar by doing `intersection(type(...), partial(...))`, but this approach feels more readable and easy to refactor.

To address the elephant in the room: I am aware that `SemiPartial` is not a *great* name - so maybe renaming the type is in order :)

It is worth noting that the method signature for `semiPartial` is 100% backwards compatible with `type` -- therefore, it would be an option to simply replace `type` with this code, and simply extend its capabilities.

It has been requested before: https://github.com/gcanti/io-ts/issues/450

There is another old but still active PoC pull request: https://github.com/gcanti/io-ts/pull/266

And it has been asked on stackoverflow a number of times (some of many examples below):
https://stackoverflow.com/questions/48230773/how-to-create-a-partial-like-that-requires-a-single-property-to-be-set
https://stackoverflow.com/questions/61311182/a-partly-partial-with-io-ts

In addition to the new functionality, the behavior of `TypeOf` is exactly as you would define an interface.  In other words the typechecker will treat the two following examples identically:

```ts
interface NameTag {
  name: string
  nickName?: string
}

const value: NameTag = {
  name: 'test',
}

const missingOptional: NameTag = {
  name: 'test',
}

const presentOptional: NameTag = {
  name: 'test',
  nickName: 'nick',
}

// has compiler error
// Property 'name' is missing in type '{ nickName: string; }' but required in type 'NameTag'.ts(2741)
const requiredMissing: NameTag = {
  nickName: 'nick'
}
```

```ts
import * as t from 'io-ts'

const NameTag = t.semiPartial(
  {
    name: t.string,
    nickName: { type: t.string, optional: true }
  }
)

type NameTag = t.TypeOf<typeof NameTag>

const missingOptional: NameTag = {
  name: 'test',
}

const presentOptional: NameTag = {
  name: 'test',
  nickName: 'nick',
}

// has compiler error
// Property 'name' is missing in type '{ nickName: string; }' but required in type '{ name: string; nickName?: string | undefined; }'.ts(2741)
const requiredMissing: NameTag = {
  nickName: 'nick'
}
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1304,6 +1304,127 @@ export function array<C extends Mixed>(item: C, name = `Array<${item.name}>`): A
   )
 }
 
+function testRequiredProps(value: { [key: string]: unknown }, count: number, keys: string[], types: Mixed[]): boolean {
+  for (let i = 0; i < count; ++i) {
+    const key = keys[i]
+    const propValue = value[key]
+    if (propValue === undefined && !hasOwnProperty.call(value, key)) {
+      return false
+    }
+    const type = types[i]
+    if (!type.is(propValue)) {
+      return false
+    }
+  }
+  return true
+}
+
+function testOptionalProps(value: { [key: string]: unknown }, count: number, keys: string[], types: Mixed[]): boolean {
+  for (let i = 0; i < count; ++i) {
+    const key = keys[i]
+    const type = types[i]
+    const propValue = value[key]
+    if (propValue !== undefined && !type.is(propValue)) {
+      return false
+    }
+  }
+  return true
+}
+
+function validateRequiredProps<R extends { [key: string]: unknown }>(
+  record: R,
+  count: number,
+  keys: string[],
+  types: Mixed[],
+  errors: Errors,
+  context: Context
+): R {
+  let validRecord = record
+  for (let i = 0; i < count; ++i) {
+    const key = keys[i]
+    const type = types[i]
+    const propValue = record[key]
+    const result = type.validate(propValue, appendContext(context, key, type, propValue))
+    if (isLeft(result)) {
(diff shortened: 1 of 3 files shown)


kalda341:
I actually quite like SemiPartial as a name. I'm less keen on the proposed API though - personally I think something like the following looks and feels a little nicer to use:

```ts
t.semiPartial({
  requiredField: t.required(t.string),
  optionalField: t.optional(t.string),
});
```

I think the part that I dislike about your API is that optional and required fields are specified in quite different ways.
Another thing to note is that there is no implementation for the unstable API. I'm unsure about where this package is going with regards to it (there haven't been a lot of updates on the issue recently), but I'm currently using a lot of the unstable API in production.

This is great work - it's about time someone decided to tackle this long standing issue, and regardless of the API I think it will be much nicer to use than the current intersection solution.

mwilliamson:
Just in case anybody in the future stumbles across this, I had a play around with this branch and fixed a few issues, specifically:

* The semi-partial type couldn't be used with `exact()` since it didn't implement `HasProps`
* It didn't support prop types where the encoded type differed from the value type
* Unions with semi-partial types didn't work properly since they couldn't extract possible tags from the semi-partial tags
* Some prop types, such as array types, were incorrectly detected as `TypedProp`, leading to errors since they would be treated as the element type rather than an array of those elements

My (possibly buggy!) changes are here: https://github.com/mwilliamson/io-ts/tree/semi-partial

I think I'll probably ending up taking a different approach, but this branch was definitely useful as inspiration!

PR #664, opened 2022-10-06
Additional UnknownRecord.is check for `undefined` added

mlegenhausen:
This PR introduces an additional check for the `undefined` value to `UnknownRecord.is`. Background is that we at Unsplash experience a "small" amount of Chrome users where `Object.prototype.toString.call(undefined)` returns `[object Window]` instead of `[object Undefined]`.

We weren't able to reproduce this error on our side, but are in the process of rolling out a patch to all users.


gcanti:
@mlegenhausen what do you think about https://github.com/gcanti/io-ts/pull/611#issuecomment-1267490294?

mlegenhausen:
Sounds interesting, still does not explain why the string representation of `undefined` is sometimes different but it seems we can not rely on `Object.prototype.toString` at all.

On first sight `u !== null && typeof u === 'object' && !Array.isArray(u)` seems to be a valid replacement.

If you want I can change my PR accordantly?

gcanti:
> If you want I can change my PR accordantly?

Yeah, looks like we can kill two birds with one stone

mlegenhausen:
Of cause to good to be true that this a drop in replacement 😉 this has false positives for `new Number()`, `new Boolean()` and so on.

mlegenhausen:
Looking at the PR https://github.com/gcanti/io-ts/commit/231b9f58218fc5e3f64ce28a183c81afff1fb642 it seems in a previous version `new Number` was a valid `UnknownRecord` then it was changed to being invalid. Should we move this case again to being valid?

gcanti (approved):

gcanti:
Thank you @mlegenhausen, patch released

PR #666, opened 2022-10-14
Added exports field to package.json

viell-dev:
Needed to use `io-ts` in a package with `moduleResolution: NodeNext` in `tsconfig.json` so I added the exports field.

I'm currently using this fork and it's working fine.

Only `package.json` has been changed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,18 @@
   "name": "io-ts",
   "version": "2.2.19",
   "description": "TypeScript runtime type system for IO decoding/encoding",
+  "exports": {
+    ".": {
+      "types": "./lib/index.d.ts",
+      "require": "./lib/index.js",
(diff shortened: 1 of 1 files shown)


viell-dev:
I noticed there are definition files outputted into the es6 folder as well but from a cursory glance they appear identical to the once in the lib folder so I didn't end up using them.

---

```json
".": {
  "types": "./lib/index.d.ts",
  "require": "./lib/index.js",
  "import": "./es6/index.js"
},
```

is just cleaner than:

```json
".": {
  "require": {
    "types": "./lib/index.d.ts",
    "default": "./lib/index.js",
  },
  "import": {
    "types": "./es6/index.d.ts",
    "default": "./es6/index.js"
  }
},
```

PR #668, opened 2022-10-30
Add optional support to type.

colin-han:
Based on the [PR #654](https://github.com/gcanti/io-ts/pull/654) , I'm trying to make a more readable solution to support optional properties of TypeScript.

```typescript
const TypeA = t.type({
  propA: t.optional(t.string),
  propB: t.string
})
```

I define a new keyword `t.optional` to annotate a properties is optional. The above declaration will generated type like following:
```typescript
interface TypeA {
  propA?: string;
  propB: string;
}
```

I borrow all UT from the PR #654 . Thanks to the author @mjburghoffer, your UT help me so many.

BTW, I don't known why I cannot pass all test (prettier and eslint) on my local environment. If it's a problem to merge this PR. let me known, I will take more research for it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "import-path-rewrite": "github:gcanti/import-path-rewrite",
     "jest": "25.2.7",
     "mocha": "7.1.1",
-    "prettier": "2.0.2",
+    "prettier": "^2.7.1",
     "rimraf": "3.0.2",
     "ts-jest": "25.3.1",
     "ts-node": "8.8.2",
--- a/src/index.ts
+++ b/src/index.ts
@@ -265,7 +265,12 @@ export interface AnyProps {
 
 function getNameFromProps(props: Props): string {
   return Object.keys(props)
-    .map((k) => `${k}: ${props[k].name}`)
+    .map((k) => {
+      if (isOptional(props[k])) {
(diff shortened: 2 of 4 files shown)


cchepelov:
This looks good!

What are the blockers for integration?

colin-han:
@cchepelov There still has some UT is not passed. And a newer version of TypeScript is required. I guess that is why this PR is not merged into master.

I'm sorry I have not enough time and skill to fix all issue. If the feature is required for you, you can use another library I published to NPM ( [io-ts-with-optional](https://www.npmjs.com/package/io-ts-with-optional) ) as a workaround.

cchepelov:
Thank you @colin-han! 

I'll see if I can get away with no optionals for the time being. It looks like io-ts' 3.0 branch is moving a bit, too. 
Afraid I'm lacking time/skill as well to help with a fix, although I'll sure be happy if that happens

colin-han:
@gcanti It looks like the new Schema system is powerful. Do you have consider to support optional syntax like mine? The current optional syntax (`C.struct({ a: C.string, b: C.number }, { c: C.boolean })`) is a little less readable.

gcanti:
@colin-han just landed in the development branch

Issue #648, opened 2022-06-30
`PathReporter` throws when a `bigint` is part of the error messages 


jonball4:
https://github.com/gcanti/io-ts/blob/920b21ea32e82b328cd14ff4b28da72ae60db505/src/PathReporter.ts#L18

This will throw an error when a `bigint` is involved.  Could we use a `replacer` in this call to handle `bigint`?

gcanti:
@jonball4 could you please provide a repro?

jonball4:
> @gcanti  @jonball4 could you please provide a repro?

```
    interface PositiveBigIntBrand {
      readonly PositiveBigInt: unique symbol
    }

    const PositiveBigInt = t.brand(
      t.bigint,
      (n): n is t.Branded<bigint, PositiveBigIntBrand> => BigInt(0) < n,
      'PositiveBigInt',
    )

    const testC = t.type({ value: PositiveBigInt });
    const testValidation = testC.decode({ value: -10n });

    // This will throw
    const value = PathReporter.report(testValidation);

```

PR #672, opened 2022-11-02
feat: stringify bigint, addresses issue #648

jonball4:
Addresses https://github.com/gcanti/io-ts/issues/648 by adding a JSON.stringify replacer for bigint.

### DIFF
--- a/src/PathReporter.ts
+++ b/src/PathReporter.ts
@@ -15,7 +15,7 @@ function stringify(v: any): string {
     }
     return v > 0 ? 'Infinity' : '-Infinity'
   }
-  return JSON.stringify(v)
+  return JSON.stringify(v, (_key: string, value: unknown) => (typeof value === 'bigint' ? `BigInt(${value})` : value))
 }
 
 function getContextPath(context: Context): string {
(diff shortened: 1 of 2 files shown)


src/PathReporter.ts
```diff
@@ -15,7 +15,7 @@ function stringify(v: any): string {
     }
     return v > 0 ? 'Infinity' : '-Infinity'
   }
-  return JSON.stringify(v)
+  return JSON.stringify(v, (_key: string, value: unknown) => (typeof value === 'bigint' ? value.toString() : value))
```
OttoAllmendinger, line 18: this way, bigints look like exactly strings which may be confusing: when passing a bigint to a property that accepts base-10 encoded integer strings, it will say `Invalid value "10" supplied to prop` when the string "10" is valid but `BigInt(10)` isnt

jonball4: good point, this suggestion makes sense to me.  Or `${value}n` for short.

jonball4: Took your suggestion, pushed. thanks again

Issue #704, opened 2023-11-28
non-enumerable records with extra keys do not pass `io-ts.record.is`, contrary to TypeScript types


tgfisher4:
### Current Behavior

`io-ts.record` type does not agree with the TypeScript's `Record` type, which AFAICT is its express goal. The discrepancy arises specifically for "non-enumerable" records, as they are called in the `io-ts` source.
In the following example (source code is also copy-pasted under **Reproducible example**) — https://stackblitz.com/edit/node-tqqi5y?file=index.ts — this behavior is unexpected:
```typescript
NonEnumerableRecord.is(hasExtraNonEnumerable)
> false
E.isRight(NonEnumerableRecord.decode(hasExtraNonEnumerable))
> false
```

### Expected behavior

In the above example, I expect
```typescript
NonEnumerableRecord.is(hasExtraNonEnumerable)
> true
E.isRight(NonEnumerableRecord.decode(hasExtraNonEnumerable))
> true
```
since `hasExtraNonEnumerable` is assignable to `NonEnumerableRecord` at the type-level.

The latter expectation assumes that it is a general principle that `.decode` should succeed whenever `.is` does: is this correct?

### Reproducible example
https://stackblitz.com/edit/node-tqqi5y?file=index.ts
```typescript
import * as t from 'io-ts';
import * as E from 'fp-ts/Either';

// io-ts and TypeScript agree that enumerable records can have extra keys
const EnumerableRecord = t.record(
  t.keyof({ a: null, b: null }),
  t.literal('value')
);
type EnumerableRecord = t.TypeOf<typeof EnumerableRecord>;
type EnumerableWithExtra = { a: 'value'; b: 'value'; c: 'value' };
const hasExtraEnumerable = { a: 'value', b: 'value', c: 'value' } as const satisfies EnumerableWithExtra;
// Typescript considers an EnumerableWithExtra to be a valid EnumerableRecord
const extraExtendsEnumerable: EnumerableWithExtra extends EnumerableRecord
  ? true
  : false = true;
const enumerable: EnumerableRecord = hasExtraEnumerable;
// so does io-ts
console.log(
  `${JSON.stringify(hasExtraEnumerable)} ${
    EnumerableRecord.is(hasExtraEnumerable) ? 'is' : 'is not'
  } ${EnumerableRecord.name}`
);
console.log(
  `${JSON.stringify(hasExtraEnumerable)} ${
[...]

PR #705, opened 2023-12-01
Handle `record` keys outside `domain`

tgfisher4:
Addresses #704.

New behavior is that:
- `record.is` allows keys outside `domain`
- `record.{en,de}code` allow but strip keys outside `domain`

Other changes:
- added/modified appropriate `record` tests
  - distinguished `HyphenatedString` from `HyphenatedStringFromNonHyphenated` so that the former can be used for (only) testing a `domain` that is a subset of `string`, and the latter for testing a prismatic `domain`
- implement `enumerableRecord` in terms of `exact`

Happy to iterate based on feedback.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -330,42 +330,17 @@ function enumerableRecord<D extends Mixed, C extends Mixed>(
   name = `{ [K in ${domain.name}]: ${codomain.name} }`
 ): RecordC<D, C> {
   const len = keys.length
+  const props: Props = {}
+  for (let i = 0; i < len; i++) {
+    props[keys[i]] = codomain
+  }
+  const exactCodec = exact(type(props, name), name)
+
   return new DictionaryType(
     name,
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -330,42 +330,17 @@ function enumerableRecord<D extends Mixed, C extends Mixed>(
   name = `{ [K in ${domain.name}]: ${codomain.name} }`
 ): RecordC<D, C> {
   const len = keys.length
+  const props: Props = {}
+  for (let i = 0; i < len; i++) {
+    props[keys[i]] = codomain
+  }
+  const exactCodec = exact(type(props, name), name)
```
gcanti, line 337: we can use `strict` here:

```ts
const exactCodec = strict(props, name)
```

gcanti:
@tgfisher4 LGTM

gcanti:
Thanks @tgfisher4

PR #719, opened 2024-12-08
Test for `readonly` type when optimizing tagged unions

polgfred:
Currently, unions of readonly types aren't optimized as tagged unions because the `getTags` function doesn't check for them. This just passes readonly's through to the underlying type.


polgfred:
@gcanti Thanks!

